PR #1, opened 2015-04-03
Added missing length check for strcpy(ifr.ifr_name, argv[x]).

snewz:


hartkopp:
Hi Sven,
thanks for your contribution! But instead of creating 'more' obviously duplicated code I would suggest to change the code to use if_nametoindex() instead of the open coded ioctl() stuff.
You can see examples of the if_nametoindex() usage in cangw.c and isotpperf.c .
Btw. a proper check of a negative return check is missing there too and needs to be added before.
Are you interested to provide a patch for this rework?
Best regards,
Oliver

snewz:
Hi Oliver,
your proposal looks nice. I'm going to pull a patch soon.
Kind regards,
Sven

PR #10, opened 2015-10-13
slcan: added listen-only flag

faryon93:
slcan_attach: use -l flag to send a L\r instead of O\r to use the can adapters listen-only mode

### DIFF
--- a/slcan_attach.c
+++ b/slcan_attach.c
@@ -58,6 +58,7 @@ void print_usage(char *prg)
 {
 	fprintf(stderr, "\nUsage: %s [options] tty\n\n", prg);
 	fprintf(stderr, "Options: -o         (send open command 'O\\r')\n");
+	fprintf(stderr, "         -l         (send listen only command 'L\\r', overrides -o)\n");
 	fprintf(stderr, "         -c         (send close command 'C\\r')\n");
 	fprintf(stderr, "         -f         (read status flags with 'F\\r' to reset error states)\n");
 	fprintf(stderr, "         -s <speed> (set CAN speed 0..8)\n");
@@ -81,6 +82,7 @@ int main(int argc, char **argv)
 	int detach = 0;
 	int waitkey = 0;
 	int send_open = 0;
+	int send_listen = 0;
 	int send_close = 0;
 	int send_read_status_flags = 0;
 	char *speed = NULL;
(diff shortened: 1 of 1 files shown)


slcan_attach.c
```diff
@@ -168,7 +170,15 @@ int main(int argc, char **argv)
...
 
-		if (send_open) {
+		/* prefer listen-only over normal open */
+		if (send_listen) {
+			sprintf(buf, "L\r");
+			write(fd, buf, strlen(buf));
+
+			if (send_open)
+				printf("ignoring -o, due to listen-only option -l\n");
+
```
marckleinebudde: Nitpick please remove this extra line. If there are no further objections, I'll do this while applying the patch.

faryon93:
okay i removed the extra line

marckleinebudde:
tnx - let's wait for Oliver's OK.

hartkopp:
Hi,

generally a good idea.

But I would suggest to put the 'listen' option directly behind the 'open' option like in the help text:

->	 while ((opt = getopt(argc, argv, "dwolcfs:b:n:?")) != -1) {
		switch (opt) {
		case 'd':
			detach = 1;
			break;

		case 'w':
			waitkey = 1;
			break;

		case 'o':
			send_open = 1;
			break;

->		case 'l':
->			send_listen = 1;
->			break;

And I would remove the

			if (send_open)
				printf("ignoring -o, due to listen-only option -l\n");

When users select '-l' they know what they are doing.
The comment that 'listen' overrides 'open' is sufficient IMO.

Regards,
Oliver

faryon93:
Hi Oliver,
thanks for your remarks. I just pushed changes which should correct the things you noted.

With best wishes,
Maxi

marckleinebudde:
Looks good to me. Oliver, should I add your Acked-by?

marckleinebudde:
Maximilian, can I add your:

    Signed-off-by: Maximilian Pachl <m@ximilian.info>

faryon93:
Hi Marck,
thanks for your reply.
sure you can add that to your commit message :)

Kind regards,
Maxi

hartkopp:
Hi all,

yes please add my acked-by.

Thanks Maxi!

marckleinebudde:
pushed to master.
tnx

PR #16, opened 2016-03-21
cangen: reset frame flags at the start of iteration

rshanmu:
This patch fixes a bug in cangen.c

Bug: When used with -m option, if a CAN FD frame with BRS option set is
sent, that flag never gets cleared for the subsequent frames.

Fix: Reset frame.flags at the start of iteration.

Signed-off-by: Ramesh Shanmugasundaram <ramesh.shanmugasundaram@bp.renesas.com>

### DIFF
--- a/cangen.c
+++ b/cangen.c
@@ -357,6 +357,7 @@ int main(int argc, char **argv)
 	}
 
 	while (running) {
+		frame.flags = 0;
 
 		if (count && (--count == 0))
 			running = 0;


hartkopp:
This is not a bug. The option '-m' makes a mix of several options (including brs). When you additionally set the '-b' option (which is NOT intended) this flag supersedes the mixed bits.
So where is the bug then? A solution could be to throw an error when options are provided that are modified by the '-m' option - or give a warning on that mis-usage ...

Do you think this is needed after my explanation?

rshanmu:
Hi Oliver,

May be bug description is not clear. Let's say -m opton is set.

```
485                 if (mix) {
486                         i = random();
487                         extended = i&1;
488                         canfd = i&2;
489                         if (canfd)
490                                 brs = i&4;
491                         rtr_frame = ((i&24) == 24); /* reduce RTR frames to 1/4 */
492                 }

```
The above code in while(running) loop, will set the canfd & brs flag based on random(). Let's say both are set. This this code will set the frame.flags with BRS option

```
364                 if (canfd){
365                         mtu = CANFD_MTU;
366                         maxdlen = CANFD_MAX_DLEN;
367                         if (brs)
368                                 frame.flags |= CANFD_BRS;

```

On the next iteration let's say "canfd" flag is set but brs is not. But the frame.flags BRS option set in previous iteration is not cleared. So the bug is, even though BRS option is not set for that iteration the frame.flags never clears BRS flags and every frame generated after this will have this flag.

This pull request fixes this bug.

hartkopp:
Thanks for the explanation & the fix.

I needed to take a look at the entire while() statement - you are definitely right!

Thanks,
Oliver

PR #20, opened 2016-05-20
added listen-only to slcand

interlocuteur:
listen-only option (-l) has been added to slcan-attach in oct 2015.
It is now replicated in slcand


hartkopp:
Applied to can-utils.

Thanks!

PR #24, opened 2016-06-13
header-guard for lib.h

ajneu:
Ah... I was trying to include lib.h from 2 different headers and ... just noticed...

A header guard would be nice as well. Also including `stdio.h` to handle `FILE` ([ref](https://github.com/linux-can/can-utils/blob/ef2ea8fa59e9123c496a109b2b4cb62dd1e3e7fa/lib.h#L145))

(Thanks for quick previous merge.)

### DIFF
--- a/lib.h
+++ b/lib.h
@@ -41,6 +41,11 @@
  *
  */
 
+#ifndef CAN_UTILS_LIB_H
+#define CAN_UTILS_LIB_H
+
+#include <stdio.h>
+
 /* buffer sizes for CAN frame string representations */
 
 #define CL_ID (sizeof("12345678##1"))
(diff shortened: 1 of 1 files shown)


ajneu:
Thanks for pulling.
Regarding a different project: libsocketcan...

I have a few changes for libsocketcan also: see https://github.com/ajneu/libsocketcan
Unfortunately libsocketcan is not hosted on github (instead [here](http://git.pengutronix.de/?p=tools/libsocketcan.git)), so I cannot send a pull-request.

(I actually sent you a mail about this 6 days ago. Did you receive it, or did it land in spam?)

PS: see also http://thread.gmane.org/gmane.linux.can/9546

PR #25, opened 2016-06-28
LGPL bittiming calculation

toonst:
All copyright holders agreed to change the license of this file to LGPLv2 so the bittiming calculation function can be used other LGPL licensed open source projects.


toonst:
There's some other kernel imports, so this new license doesn't apply.

Issue #26, opened 2016-08-30
candump exits when device goes down


ajneu:
Hi there,

`candump` is a useful tool, but the fact that
```bash
candump vcan0
```
will just exit and print `read: Network is down`, when I pull the device down (via `sudo ip link set vcan0 down`) is unfortunate.

PS: did you know, that I do **not** get the message `read: Network is down` and exit, when using `any` as device?? ;) Unfortunately using `any`... does not report on down/up status though.

Minor aside:
There's a common use-case for binaries using [libsocketcan](http://git.pengutronix.de/?p=tools/libsocketcan.git;a=summary), to pull the device down (`can_do_stop`) and back up again (`can_do_start`). It's the only way I know of, to clear the socketcan transmit buffer. (And I really want to clear pending messages from it, because if a CAN-bus reconnect only occurs in say 5 minutes, then I don't want to transmit 5-minute old frames on reconnection.)

To try to work around the binary just exiting, I scripted up something, where I use `ip monitor` to detect when a device goes down and back up, and then invoke `candump` again. Try:
```bash
ip monitor link | while IFS=':' read num name status; do
    num=$(echo    $num    | xargs) # strip whitespace
    name=$(echo   $name   | xargs) # strip whitespace
    status=$(echo $status | xargs) # strip whitespace
    echo "   num   ==$num"
    echo "   name  ==$name"
    echo "   status==$status"
    read extraline
    ## ...
    ## ... code to handle change of status from DOWN to UNKOWN
    ## ...
done
```
But it's an ugly solution, because that script is so slow that the immediate frames sent when the device is back up *are just missed*. That's not ok.

So I decided to modify the `candump.c` code.
I've got 2 versions:
* https://github.com/ajneu/can-utils/tree/no_exit_on_down -- extremely simple
* https://github.com/ajneu/can-utils/tree/down_up_reporting -- quite nice, since I report when a device goes up or down

I believe it would be nice for something like the bottom one to go into upstream here.
I'm using some code (function `send_dump_request`) from [libsocketcan.c](http://git.pengutronix.de/?p=tools/libsocketcan.git;a=blob;f=src/libsocketcan.c;hb=HEAD). I think it's compatible with the very permissive Volkswagen license of `candump.c`.

Please clone the bottom one ([down_up_reporting](https://github.com/ajneu/can-utils/tree/down_up_reporting)) and give it a try: `candump -R -D vcan0`. Bring down devices from a second shell and bring them up again. What do you think?

[...]

PR #29, opened 2016-08-31
candump down/up device monitoring

ajneu:
See #26

### DIFF
--- a/candump.c
+++ b/candump.c
@@ -61,6 +61,7 @@
 
 #include <linux/can.h>
 #include <linux/can/raw.h>
+#include <linux/rtnetlink.h>
 
 #include "terminal.h"
 #include "lib.h"
@@ -120,6 +121,8 @@ void print_usage(char *prg)
 	fprintf(stderr, "         -L          (use log file format on stdout)\n");
 	fprintf(stderr, "         -n <count>  (terminate after receiption of <count> CAN frames)\n");
 	fprintf(stderr, "         -r <size>   (set socket receive buffer to <size>)\n");
+	fprintf(stderr, "         -R          (Report down-and-up-again status for \"detected\" devices.\n");
+	fprintf(stderr, "                      For any device that is only matched via \"any\": Such a device is only \"detected\", once you have received at least one frame from that device.\n");
 	fprintf(stderr, "         -D          (Don't exit if a \"detected\" can device goes down.\n");
 	fprintf(stderr, "         -d          (monitor dropped CAN frames)\n");
 	fprintf(stderr, "         -e          (dump CAN error frames in human-readable format)\n");
@@ -150,6 +153,19 @@ void print_usage(char *prg)
 void sigterm(int signo)
 {
 	running = 0;
+}
+
+// idx of "detected" devices is cached
+int is_idx_cached(int ifidx) {
+	int cached = 0;
+	int i;
+	for (i=0; i < MAXIFNAMES; i++) {
+		if (dindex[i] == ifidx) {
+			cached = 1;
+			break;
+		}
+	}
+	return cached;
 }
 
 int idx2dindex(int ifidx, int socket) {
@@ -201,6 +217,32 @@ int idx2dindex(int ifidx, int socket) {
 	return i;
 }
 
+// http://git.pengutronix.de/?p=tools/libsocketcan.git;a=blob;f=src/libsocketcan.c;hb=HEAD
+static int send_dump_request(int fd, int family, int type)
+{
+    struct get_req {
+	struct nlmsghdr n;
+	struct rtgenmsg g;
+    } req;
+
+    memset(&req, 0, sizeof(req));
+
+    req.n.nlmsg_len = sizeof(req);
+    req.n.nlmsg_type = type;
+    req.n.nlmsg_flags = NLM_F_REQUEST | NLM_F_ROOT | NLM_F_MATCH;
+    req.n.nlmsg_pid = 0;
+    req.n.nlmsg_seq = 0;
+
+    req.g.rtgen_family = family;
+
+    return send(fd, (void *)&req, sizeof(req), 0);
+}
+/* altern: 
+   void netlink_getlink(int nsock)   from 
+   http://centaur.sch.bme.hu/~leait/projects/openwrt/
+   http://centaur.sch.bme.hu/~leait/projects/openwrt/files/sample.c
+*/
+
 int main(int argc, char **argv)
 {
 	fd_set rdfs;
@@ -209,6 +251,7 @@ int main(int argc, char **argv)
 	useconds_t bridge_delay = 0;
 	unsigned char timestamp = 0;
 	unsigned char down_causes_exit = 1;
+	unsigned char report_down_up = 0;
 	unsigned char dropmonitor = 0;
 	unsigned char extra_msg_info = 0;
 	unsigned char silent = SILENT_INI;
@@ -236,6 +279,10 @@ int main(int argc, char **argv)
 	struct timeval tv, last_tv;
 	struct timeval timeout, timeout_config = { 0, 0 }, *timeout_current = NULL;
 	FILE *logfile = NULL;
+	struct netlink_struc {
+		int s;
+		struct sockaddr_nl saddr;
+	} netlink_s;
 
 	signal(SIGTERM, sigterm);
 	signal(SIGHUP, sigterm);
@@ -244,7 +291,7 @@ int main(int argc, char **argv)
 	last_tv.tv_sec  = 0;
 	last_tv.tv_usec = 0;
 
-	while ((opt = getopt(argc, argv, "t:ciaSs:b:B:u:lDdxLn:r:heT:?")) != -1) {
+	while ((opt = getopt(argc, argv, "t:ciaSs:b:B:u:lDdxLn:Rr:heT:?")) != -1) {
 		switch (opt) {
 		case 't':
 			timestamp = optarg[0];
@@ -363,6 +410,10 @@ int main(int argc, char **argv)
 			}
 			break;
 
+		case 'R':
+			report_down_up = 1;
+			break;
+
 		case 'T':
 			errno = 0;
 			timeout_config.tv_usec = strtol(optarg, NULL, 0);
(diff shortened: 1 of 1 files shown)


ajneu:
To test: 
```bash
candump -R -D vcan0
```

In other shell change can-device down and up...:
```bash
sudo ip link set vcan0 down
sudo ip link set vcan0 up
sudo ip link set vcan0 down
sudo ip link set vcan0 up
# etc...
```

candump.c
```diff
@@ -152,6 +155,20 @@ void sigterm(int signo)
 	running = 0;
 }
 
+// idx of "detected" devices is cached
+int is_idx_cached(int ifidx) {
+	int cached = 0;
+	int i;
+
+	for (i=0; i < MAXIFNAMES; i++) {
```
marckleinebudde: space around "="

candump.c
```diff
@@ -236,6 +281,16 @@ int main(int argc, char **argv)
 	struct timeval tv, last_tv;
 	struct timeval timeout, timeout_config = { 0, 0 }, *timeout_current = NULL;
 	FILE *logfile = NULL;
+	struct netlink_struc {
+		int s;
+		struct sockaddr_nl saddr;
+	} netlink_s = {
+			.s = socket(AF_NETLINK, SOCK_RAW, NETLINK_ROUTE),
```
marckleinebudde: Sorry for not beeing clear in the first review, better use only C99 initializers for constant types. So move the: `s = socket();` next to the `bind()`.

candump.c
```diff
@@ -399,9 +458,17 @@ int main(int argc, char **argv)
 			silent = SILENT_OFF; /* default output */
 	}
 
+	// netlink to determine down/up
+	bind(netlink_s.s, (struct sockaddr *) &netlink_s.saddr, sizeof(netlink_s.saddr));
```
marckleinebudde: please add error handling

marckleinebudde: remove the space between the cast and the &netlink_s

candump.c
```diff
@@ -635,6 +704,41 @@ int main(int argc, char **argv)
 			continue;
 		}
 
+		// check netlink socket
+		if (FD_ISSET(netlink_s.s, &rdfs)) {
+			char buf[4096];
```
marckleinebudde: please move this into a separate function

candump.c
```diff
@@ -635,6 +704,41 @@ int main(int argc, char **argv)
...
+				if (nh->nlmsg_type == NLMSG_ERROR)
+					continue;
+				if (nh->nlmsg_type == RTM_NEWLINK) {
+					struct ifinfomsg *rtif = NLMSG_DATA(nh);
+
+					if (is_idx_cached(rtif->ifi_index)) {
+						char ifname[IF_NAMESIZE];
+						if_indextoname(rtif->ifi_index, ifname);
+
+						const int will_exit = (down_causes_exit && !(rtif->ifi_flags&IFF_RUNNING));
```
marckleinebudde: please move all variable declarations to the beginning of the scope. (I know, kernel C-style is not consequent in using C99 or not :)

candump.c
```diff
@@ -647,16 +751,23 @@ int main(int argc, char **argv)
...
 
+				idx = idx2dindex(addr.can_ifindex, s[i]);
+
 				nbytes = recvmsg(s[i], &msg, 0);
 				if (nbytes < 0) {
-					if ((errno == ENETDOWN) && !down_causes_exit) {
-						fprintf(stderr, "read: %s \t\t(not exiting because of -D flag)\n", strerror(errno));
+					if (errno == ENETDOWN) {
+					    if (down_causes_exit) {
+						//if (!report_down_up) { /* no previous reporting message, so report here before exit */
```
marckleinebudde: If you don't need the commented out stuff, remove it

hartkopp:
I generally don't like this patch.
IMO it can make sense to omit the exit of candump as the previous implemented.
(And when doing so the output should not look like "read: Network is down 		(not exiting because of -D flag)" but "vcan0: interface gone down (temporarily?)")
But IMO this patch is just bloating the code. In which use-case does it really help?
Aka: Why should people invoke an ifdown/ifup at runtime?
And what happens when you completely remove the interface?
Do you probably want to invoke "ip link set can0 restart" instead??

candump.c
```diff
@@ -647,16 +751,23 @@ int main(int argc, char **argv)
 				msg.msg_controllen = sizeof(ctrlmsg);  
 				msg.msg_flags = 0;
 
+				idx = idx2dindex(addr.can_ifindex, s[i]);
```
hartkopp: Additionally to the fact that I don't like the patch addr.can_ifindex is only valid AFTER recvmsg() ... :-)

ajneu:
> But IMO this patch is just bloating the code. In which use-case does it really help?

There's a common use-case for binaries using [libsocketcan](http://git.pengutronix.de/?p=tools/libsocketcan.git;a=summary), to pull the device down (`can_do_stop`) and back up again (`can_do_start`). It's the only way I know of, to clear the socketcan transmit buffer. (And I really want to clear pending messages from it, because if a CAN-bus reconnect only occurs in say 5 minutes, then I don't want to transmit 5-minute old frames on reconnection.) See also #26 

> Aka: Why should people invoke an ifdown/ifup at runtime?

Answered above.

> And what happens when you completely remove the interface?

Does it have a usecase?

Here's what happens:
```bash
   $./candump -D -R vcan0
# vcan0 is up

                      ## In other shell do:     sudo ip link delete vcan0
# vcan0 is down
read: No such device
```

> Do you probably want to invoke "ip link set can0 restart" instead??

Does that command exist???
```bash
   $ ip -V
ip utility, iproute2-ss160518   ## it's this one:  https://packages.debian.org/stretch/iproute2

   $ ip link help 2>&1 | grep restart
   ## no result

   $ sudo ip link set vcan0 restart
Error: either "dev" is duplicate, or "restart" is a garbage.
```

PS: But I've now rechecked libsocketcan, and found that I missed `can_do_restart` till now (I previously always did `can_do_stop(can0); can_do_start(can0)`. It seems that libsocketcan's `can_do_restart(vcan0)` does not cause down-and-up... so either it is buggy, or it does some kind of transparent restart (which would be great.). Does `can_do_restart(can0)` clear the socketcan transmit buffer?

candump.c
ajneu: I've changed it, but I'm not sure if it was, what you were aiming at...

ajneu:
> But IMO this patch is just bloating the code.

Well I've now move the netlink parts into a separate function as suggested by Marc Kleine-Budde [here](https://github.com/linux-can/can-utils/pull/29#discussion-diff-77008181).
'hope that makes it somewhat better.

ajneu:
@ Marc Kleine-Budde: I'm interested in suggesting some changes to upstream libsocketcan ([ref](https://github.com/ajneu/libsocketcan/commits/master)). What is the correct way of proposing changes for that project?
Thanks

hartkopp:
> >  But IMO this patch is just bloating the code. In which use-case does it really help?

> There's a common use-case for binaries using libsocketcan, to pull the device down (can_do_stop) and back up again (can_do_start). It's the only way I know of, to clear the socketcan transmit buffer. (And I really want to clear pending messages from it, because if a CAN-bus reconnect only occurs in say 5 minutes, then I don't want to transmit 5-minute old frames on reconnection.) See also #26 

Ok. This is a use-case for doing a down/up sequence.

But do we really need to have this blabla of 'is down' and 'is up' in candump for this?
Having the information of a 'down' interface seems to be enough - as it is a new reaction instead of terminating candump. But when the interface comes up again some traffic will show up anyway.

Two additional infos:

1. 'ip link set can0 type can restart' is a way to reset the CAN controller in the case of BUS_OFF
When you define 'ip link set can0 type can restart-ms 100' the controller is restarted 100ms after BUS_OFF automatically. So it does not really help in the case of an open bus without a second node.
And I saw that 'ip link set can0 type can restart' says 'RTNETLINK answers: Device or resource busy' - hm. I thought this should work ...

2. I found this interesting document and comment https://www.coverfire.com/articles/queueing-in-the-linux-network-stack/#comment-480695 on the Internet. But even if we flush the tx queue we still would have the single CAN frame in the CAN controller.

We could think about a new  'ip link set can0 type can flush' command to flush the tx path without doing a down/up sequence.

ajneu:
> But do we really need to have this blabla of 'is down' and 'is up' in candump for this?

Yes, I want it, because then I can see if a device is pulled down and up again. It's only active with `-R flag`.

> We could think about a new 'ip link set can0 type can flush' command to flush the tx path without doing a down/up sequence.
Yes. 
(But I'll be using down/up, since it's working in my current kernel, and I don't have to mess around with patching the kernel, etc. 
For me, a patched candump is simpler... and hope that my patch here can be accepted upstream.)

hartkopp:
> > But do we really need to have this blabla of 'is down' and 'is up' in candump for this?

> Yes, I want it, because then I can see if a device is pulled down and up again. It's only active with -R flag.

Ok - I must admit that my question was a rhetorical one.

The idea behind Unix tools is its philosophy https://en.wikipedia.org/wiki/Unix_philosophy :
Make each program do one thing well. To do a new job, build afresh rather than complicate old programs by adding new "features".

candump is a tool to display or dump CAN data.
candump is no tool to monitor and display CAN interface changes.

I'm fine with your change to preserve candump's termination so that it behaves with explicitly named CAN interfaces in the same way as it does with the 'any' device.
But adding all these netlink stuff just to print changing interface states on stderr? No.

Following the Unix philosophy you may create a new tool that displays CAN interface states - e.g. in another terminal of your system. Or you can send CAN interface states on a virtual CAN interface and dump this interface with candump too.

Or you may have your desired functionality in your private repo.
But don't expect this netlink stuff to become part of candump.

ajneu:
> Following the Unix philosophy you may create a new tool that displays CAN interface states - e.g. in another terminal of your system. 

I want an output that preserves temporal order of occurence to 100%.
How can I combine candump with CAN interface states?

Perhaps you though with the following, but I see it as rather obscure:

> Or you can send CAN interface states on a virtual CAN interface and dump this interface with candump too.

I thought the Unix tools and philosophy work through file-writing and file-reading. Dumping via a virtual CAN interface seems a bit far-fetched, to be honest.
Do you mean that I should invent some CAN-frame ID's and payload data to symbolize up/down etc?
That's about as far away from elegant, as I can conceive. And I'm not sure I'll have preserved temporal order.

> Or you may have your desired functionality in your private repo.

Yip. This is the only viable solution to my needs as user. (Ponder that...)

hartkopp:
>> Or you may have your desired functionality in your private repo.

> Yip. This is the only viable solution to my needs as user. (Ponder that...)

I did and still do.

candump receives data from the CAN bus.
When you shut down the interface you now get an information about it.
When the interface comes up again you get new CAN frames.

So where's the difference between getting new CAN frames AND getting new CAN frames and know that the interface has gone up 543 ms before?

ajneu:
> So where's the difference between getting new CAN frames AND getting new CAN frames and know that the interface has gone up 543 ms before?

The difference is that in the second case I know that the interface was down and has gone up.

Why do I want this information? Because then when I do long-term tests, with candump, I can look at the log and know that libsocketcan's `can_do_stop` and `can_do_start` (which I happen to use) were called and when they were called. Otherwise I'm just guessing (or must fiddle with my user application, and turn it into a tool, that I expect candump to be)...

hartkopp:
> Because then when I do long-term tests, with candump, I can look at the log and know that libsocketcan's `can_do_stop` and `can_do_start` (which I happen to use) were called and when they were called. Otherwise I'm just guessing (or must fiddle with my user application, and turn it into a tool, that I expect candump to be)...

Yes. Fiddle it into **your** user application and print a proper timestamp when you perform the `can_do_stop` and `can_do_start` actions with it.
https://memegenerator.net/instance/71557527

PR #31, opened 2016-10-04
slcand: log to stdout when running in foreground

umlaeute:
when slcand runs in daemon mode, it makes sense to output via syslog.
but when running in foreground, it is nicer to have all the info on the stdout.

### DIFF
--- a/slcand.c
+++ b/slcand.c
@@ -39,6 +39,7 @@
 #include <termios.h>
 #include <linux/tty.h>
 #include <linux/sockios.h>
+#include <stdarg.h>
 
 /* Change this to whatever your daemon is called */
 #define DAEMON_NAME "slcand"
@@ -53,6 +54,19 @@
 #define FLOW_NONE 0
 #define FLOW_HW 1
 #define FLOW_SW 2
+
+void fake_syslog(int priority, const char*format, ...)
+{
+	va_list ap;
+	printf("[%d] ", priority);
+	va_start(ap, format);
+	vprintf(format, ap);
+	va_end(ap);
+	printf("\n");
+}
+typedef void (*syslog_t)(int priority, const char*format, ...);
+syslog_t syslogger;
+
 
 void print_usage(char *prg)
 {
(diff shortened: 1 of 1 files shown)


slcand.c
```diff
@@ -54,6 +55,19 @@
...
+void fake_syslog(int priority, const char*format, ...)
+{
+	va_list ap;
+	printf("[%d] ", priority);
+	va_start(ap, format);
+	vprintf(format, ap);
+	va_end(ap);
+	printf("\n");
+}
+typedef void (*syslog_t)(int priority, const char*format, ...);
```
marckleinebudde: Please add a space before the *: `const char *format, ...);`

slcand.c
```diff
@@ -54,6 +55,19 @@
...
+syslog_t syslogger;
+
+
```
marckleinebudde: one newline should be enough here

slcand.c
```diff
@@ -54,6 +55,19 @@
 #define FLOW_HW 1
 #define FLOW_SW 2
 
+void fake_syslog(int priority, const char*format, ...)
+{
+	va_list ap;
```
marckleinebudde: one newline here, please

slcand.c
```diff
@@ -185,6 +199,7 @@ int main(int argc, char *argv[])
 	int fd;
 
 	ttypath[0] = '\0';
+	syslogger = syslog;
```
marckleinebudde: you can move this to the declaration of `syslogger`

slcand.c
```diff
@@ -54,6 +55,19 @@
...
+{
+	va_list ap;
+	printf("[%d] ", priority);
+	va_start(ap, format);
+	vprintf(format, ap);
+	va_end(ap);
+	printf("\n");
+}
+typedef void (*syslog_t)(int priority, const char*format, ...);
+syslog_t syslogger;
```
marckleinebudde: make it static please

marckleinebudde:
Squashed and pushed to master. tnx.

umlaeute:
thanks.
i *might* have caused less trouble if there was a documentation about `CONTRIBUTING` (detailing both the code style (<kbd>tab</kbd> indentention!) and the commit style (squash).

(though to be honest, it might as well not)

PR #35, opened 2016-12-19
bcmserver: fix ifr_name scan size

jihochu:
Fix fr_name scan size to the IFNAMSIZ-1.

Signed-off-by: Jiho Chu <jiho.chu@samsung.com>

### DIFF
--- a/bcmserver.c
+++ b/bcmserver.c
@@ -150,11 +150,17 @@ int main(int argc, char **argv)
 
 	char buf[MAXLEN];
 	char rxmsg[50];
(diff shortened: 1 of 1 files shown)


hartkopp:
You should add a check whether snprintf(format, MAXLEN, ... returns a value which is really smaller than MAXLEN to make sure, that the format string is completely written and not truncated by snprintf.

jihochu:
Hi,
Thanks for your check. I modified it.

hartkopp:
Applied with a more informative commit message.

Thanks!

PR #37, opened 2017-02-13
isotptun: daemonize by default, add -F for old behavior

kohtala:
We want isotptun to exit when the tun device can be used. This helps system
startup continue configuration and use of the interface at correct time.

Log errors to syslog when daemonized or to stderr when in foreground.

Exit consistently with EXIT_FAILURE when the tun device was not set up as
defined on the command line.

Signed-off-by: Marko Kohtala <marko.kohtala@gmail.com>

### DIFF
--- a/isotptun.c
+++ b/isotptun.c
@@ -55,6 +55,8 @@
 #include <libgen.h>
 #include <errno.h>
 #include <signal.h>
+#include <stdarg.h>
+#include <syslog.h>
 
 #include <net/if.h>
 #include <sys/types.h>
@@ -65,6 +67,9 @@
 #include <linux/can/isotp.h>
 #include <linux/if_tun.h>
 
+/* Change this to whatever your daemon is called */
+#define DAEMON_NAME "isotptun"
+
 #define NO_CAN_ID 0xFFFFFFFFU
 #define DEFAULT_NAME "ctun%d"
 
@@ -73,6 +78,26 @@
 #define BUF_LEN (MAX_PDU_LENGTH + 1)
 
 static volatile int running = 1;
+
+static void fake_syslog(int priority, const char *format, ...)
+{
+	va_list ap;
+
+	fprintf(stderr, "[%d] ", priority);
+	va_start(ap, format);
+	vfprintf(stderr, format, ap);
+	va_end(ap);
+	fprintf(stderr, "\n");
+}
+
+typedef void (*syslog_t)(int priority, const char *format, ...);
+static syslog_t syslogger = syslog;
+
+void perror_syslog(const char *s)
+{
+	const char *colon = s ? ": " : "";
+	syslogger(LOG_ERR, "%s%s%s", s, colon, strerror(errno));
+}
 
 void print_usage(char *prg)
 {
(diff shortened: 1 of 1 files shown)


isotptun.c
```diff
@@ -154,9 +177,9 @@ int main(int argc, char **argv)
 			else if (elements == 2)
 				opts.flags |= (CAN_ISOTP_EXTEND_ADDR | CAN_ISOTP_RX_EXT_ADDR);
 			else {
-				printf("incorrect extended addr values '%s'.\n", optarg);
+				fprintf(stderr, "incorrect extended addr values '%s'.\n", optarg);
 				print_usage(basename(argv[0]));
-				exit(0);
+				exit(EXIT_FAILURE);
 			}
```
marckleinebudde: Can you move all exit(EXIT_FAILURE) related changes into a single commit - preferred before you add daemon mode.

isotptun.c
```diff
@@ -118,14 +144,11 @@ int main(int argc, char **argv)
 	unsigned char buffer[BUF_LEN];
 	static char name[IFNAMSIZ] = DEFAULT_NAME;
 	int nbytes;
-
-	signal(SIGTERM, sigterm);
-	signal(SIGHUP, sigterm);
-	signal(SIGINT, sigterm);
+	int run_as_daemon = 1;
```
marckleinebudde: bool

isotptun.c
kohtala: Since the use of it as a daemon requires changing some of the earlier success codes (0) to failures (EXIT_FAILURE) I thought they go logically into the same commit. Whoever is starting the daemon needs to know if it the daemon has started or if there was some error preventing it from starting. When it was only foreground, the exit code did not matter so much.

marckleinebudde: On the other hand it's better to just exit with an error, if there is one.

kohtala:
I took parts from slcand. slcand is by default a daemon and has -F option to keep it in foreground. So this is for consistency between tools. Thinking about it now, it would be possible to rename isotptun to isotptund for added consistency and to indicate the change. Would looking for 'd' at the and of argv[0] for a default behavior be too much? Could then add -B (or I'd prefer -D for daemon) for when it is not named as daemon.

isotptun.c
kohtala: This is from slcand. I now looked, and there is little use of `bool`. Even isotptun.c already at least variables `running` and `verbose` defined as `int`, so use of `bool` would be inconsistent.

marckleinebudde:
Yes, it's inconsistent, but I'd rather not change the default behaviour. I think argv[0] magic is a bit too much :)

isotptun.c
marckleinebudde: Ok, then make it an `int`. Some tools use several levels of verbosity, this is why `verbose` might be an `int`.

isotptun.c
kohtala: Would you still prefer the changes to EXIT_FAILURE in separate commit?

marckleinebudde: The change on error -> `exit(EXIT_FAILURE)` and the non functional change `exit(0)` to `exit(EXIT_SUCCESS)` can go into a single patch.

kohtala: I'm a tad bit uncertain I get this right. Single patch separate from the daemon feature? Can do like that as well. Clear communication is so difficult... :-)

Then there are the changes to print out errors to stderr instead of stdout. Opinion on them?

marckleinebudde: Put _all_ changes regarding `exit(...)` into the first patch. All remaining changes into the second.

kohtala: Ah. You are so fast in replying. Already made separate commits for the stderr.

marckleinebudde: Ok, keep it seperate commits then - More is usually better. :D

kohtala:
Is there anything more I should do for this?

If anything else is needed from me, I'd have an interest in doing it now or some time next week. It'll be more difficult to come back to this later.

PR #48, opened 2017-08-08
use a macro for standardization

darbedar:
I was reading the code to learn more about can but I noticed that the hex_asc_upper_lo macro which is defined before do the same thing as  hex_asc_upper[id & 0xF] so it needs to change for standardization


hartkopp:
Good catch!
Thanks!

PR #59, opened 2017-12-02
Add option to read hardware timestamps.

FabianInostroza:
I took the code from https://marc.info/?l=linux-can&m=138244457221323&w=2.

Unfortunately only the pcan driver provides hardware timestamps.

Note that I don't own a PCAN adapter, I tested this modification using a self made pcan compatible adapter, receiving one message every 400 us and the hardware timestamp sometimes jumps back or forward. The same self made adapter gives correct timestamps in Windows with the official driver and PCAN-view.

### DIFF
--- a/candump.c
+++ b/candump.c
@@ -58,6 +58,7 @@
 #include <sys/ioctl.h>
 #include <sys/uio.h>
 #include <net/if.h>
+#include <linux/net_tstamp.h>
 
 #include <linux/can.h>
 #include <linux/can/raw.h>
@@ -108,6 +109,7 @@ void print_usage(char *prg)
 	fprintf(stderr, "\nUsage: %s [options] <CAN interface>+\n", prg);
 	fprintf(stderr, "  (use CTRL-C to terminate %s)\n\n", prg);
 	fprintf(stderr, "Options: -t <type>   (timestamp: (a)bsolute/(d)elta/(z)ero/(A)bsolute w date)\n");
+	fprintf(stderr, "         -H          (Read hardware timestamps)\n");
 	fprintf(stderr, "         -c          (increment color mode level)\n");
 	fprintf(stderr, "         -i          (binary output - may exceed 80 chars/line)\n");
 	fprintf(stderr, "         -a          (enable additional ASCII output)\n");
@@ -208,6 +210,7 @@ int main(int argc, char **argv)
 	int bridge = 0;
 	useconds_t bridge_delay = 0;
 	unsigned char timestamp = 0;
+	unsigned char hwtimestamp = 0;
 	unsigned char down_causes_exit = 1;
 	unsigned char dropmonitor = 0;
 	unsigned char extra_msg_info = 0;
@@ -224,7 +227,10 @@ int main(int argc, char **argv)
 	int join_filter;
 	char *ptr, *nptr;
 	struct sockaddr_can addr;
-	char ctrlmsg[CMSG_SPACE(sizeof(struct timeval)) + CMSG_SPACE(sizeof(__u32))];
+	struct {
+		struct cmsghdr cm;
+		char control[512];
+	} ctrlmsg;
 	struct iovec iov;
 	struct msghdr msg;
 	struct cmsghdr *cmsg;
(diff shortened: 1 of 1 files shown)


candump.c
```diff
@@ -688,8 +707,16 @@ int main(int argc, char **argv)
...
 						memcpy(&tv, CMSG_DATA(cmsg), sizeof(tv));
+					} else if (cmsg->cmsg_type == SO_TIMESTAMPING) {
+
+						struct timespec *stamp = (struct timespec *)CMSG_DATA(cmsg);
+
+						stamp += 2;
+						tv.tv_sec = stamp->tv_sec;
+						tv.tv_usec = stamp->tv_nsec/1000;
+					}
 					else if (cmsg->cmsg_type == SO_RXQ_OVFL)
```
marckleinebudde: Please make it:
`} else if () {`

candump.c
```diff
@@ -688,8 +707,16 @@ int main(int argc, char **argv)
...
 				     cmsg && (cmsg->cmsg_level == SOL_SOCKET);
 				     cmsg = CMSG_NXTHDR(&msg,cmsg)) {
-					if (cmsg->cmsg_type == SO_TIMESTAMP)
+					if (cmsg->cmsg_type == SO_TIMESTAMP) {
 						memcpy(&tv, CMSG_DATA(cmsg), sizeof(tv));
+					} else if (cmsg->cmsg_type == SO_TIMESTAMPING) {
+
+						struct timespec *stamp = (struct timespec *)CMSG_DATA(cmsg);
+
+						stamp += 2;
```
hartkopp: This offset calculation might be clear when looking at the referenced code or in the Linux code.
But "stamp += 2;" should be commented and properly referenced WHY we can find the hardware timestamp at this point.

Btw. thanks for the patch :-)

FabianInostroza:
Hi, it's done.

I left candump logging to a file for a long time (15 min aprox) and the timestamp wraps at a certain point (not constant). I don't think this is related to the adapter I'm using.

FabianInostroza:
Hi,

The problem with the timestamp jumps was in the firmware, the cause of the wrap after a long time was oscillator frequency error (15 secs every 4200 sec).

Is it possible for the kernel to reorder the messages before being sent to the clients?
Now I have noticed that candump show some messages reordered, look at the following candump log 
where the first four data bytes contain a sequence number
```
(1512440012.199080) can0 1F0#7269000005060708
(1512440012.199946) can0 1F1#7369000005060708
(1512440012.199946) can0 00ABCDEF#7469000005060708 
(1512440012.201872) can0 1F0#7869000005060708 ---> sequence number jumps
(1512440012.201872) can0 1F1#7969000005060708
(1512440012.201872) can0 00ABCDEF#7A69000005060708
(1512440012.201013) can0 1F0#7569000005060708 ---> sequence number jumps
(1512440012.201014) can0 1F1#7669000005060708
```
Also look at the timestamps, they also are out of order.
I verified the usb packets that the device sends (using usbmon and wireshark) and the messages 
are sent in correct order and timestamps are correct.
Below is a capture taken at the same time using hardware timestamps
```
(0000005020.582582) can0 1F0#7269000005060708
(0000005020.582923) can0 1F1#7369000005060708
(0000005020.583307) can0 00ABCDEF#7469000005060708
(0000005020.584672) can0 1F0#7869000005060708
(0000005020.585014) can0 1F1#7969000005060708
(0000005020.585440) can0 00ABCDEF#7A69000005060708
(0000005020.583648) can0 1F0#7569000005060708
(0000005020.583990) can0 1F1#7669000005060708
```
I used the script can_logger.py provided by python-can and also show this behavior.

hartkopp:
Can you tell more about the used CAN hardware? Is it USB based?
See https://marc.info/?l=linux-can&m=148007442317274&w=2
Can we discuss this - not candump-related - issue at linux-can mailing list?
Tnx.

FabianInostroza:
Yes, it is USB based. I didn't found that thread when googling.
I will try the suggestions posted in that thread.
Do you want me to include some additional information?

hartkopp:
I personally used a PCAN USB pro FD to test your support for hardware timestamps.
You might just subscribe to the linux-can mailing list, see:
http://vger.kernel.org/vger-lists.html#linux-can 

The mailing list does not have much traffic.

FabianInostroza:
> I personally used a PCAN USB pro FD to test your support for hardware timestamps.
> You might just subscribe to the linux-can mailing list, see:
> http://vger.kernel.org/vger-lists.html#linux-can
> 
> The mailing list does not have much traffic.

@hartkopp Hi, I subscribed to the mailing list and sent a email with a patch but it isn't shown in the archives, I've received mails sent by other members but no error or notification about my own email.
What could be wrong?

hartkopp:
Did you send it with `git send-email`?
Usually only HTML mails are inappropriate and cause problems.

FabianInostroza:
> Did you send it with `git send-email`?
> Usually only HTML mails are inappropriate and cause problems.

Mmm, that could be the cause. I sent the email using gmail web client and after the mail was sent I noticed that the format of the patch was messed up and the mail was sent with some html.
Will try again using plain text.

PR #64, opened 2018-01-18
testj1939: initialize verbose variable

BastianStender:
verbose is used uninitialized when testj1939 is not called with -v.


marckleinebudde:
cherry-picked

PR #65, opened 2018-01-19
jacd: add verbosity and fix missing interface

BastianStender:


jacd.c
```diff
@@ -140,6 +141,28 @@ static int lookup_name(uint64_t name)
 
 }
 
+static const char *canaddr2str(const struct sockaddr_can *paddr)
+{
```
marckleinebudde: Is this the same or similar to https://github.com/linux-can/can-utils/blob/j1939/libj1939.c#L149 `libj1939_addr2str()`?

jacd.c
```diff
@@ -217,6 +217,7 @@ static int open_socket(const char *device, uint64_t name)
 			.addr = J1939_IDLE_ADDR,
 			.pgn = 0x0ee00,
 		},
+		.can_ifindex = if_nametoindex(s.intf),
```
marckleinebudde: This is a direct consequence of not allowing bind to any device ("3cb6a1204598 j1939: socket: bind() cleanups"), i.e. `can_ifindex == 0`. This means before this change `jacd` would receive from any interface that was opened by another program.

jacd.c
BastianStender: Yes, exactly the same. We could move that into a separate file, maybe "canhelper" or something similar?

marckleinebudde: What about `libj1939.c` :)

jacd.c
BastianStender: Sounds reasonable, what should we do here?

marckleinebudde: Just wanted to explain why this is necessary now and that I broke it with my commit.

jacd.c
BastianStender: Argh, sorry, I misread that. I actually took it from testj1939, will replace it in this PR as well as in testj1939. Yay, less duplicate code :)

marckleinebudde: 👍

jacd.c
BastianStender: Ah, got it. I will add that information to the commit message.

BastianStender:
I just pushed the requested changes:
- the commit message contains @marckleinebudde's explanation
- the verbose prints use libj1939_addr2str instead of a local helper variant providing the same functionality

PR #69, opened 2018-02-12
Reformat output of help/usage information

gerasiov:
While packaging  can-utils for Debian I finally decided to generate manpages for the tools with help2man[1].

I've added something like this to Makefile:
```
HELP2MAN = help2man
VERSION = git20180118

PROGRAMS = ...
MANPAGES = $(PROGRAMS:%=%.1)

%.1: %
        $(HELP2MAN) --no-info --no-discard-stderr --help-option="-?" --version-string="$< $(VERSION)" ./$< > $@
```
Real code could be found in debian/rules [2]

Then using help2man I generated bunch of manpages. Unfortunately most the tools outputs usage information if format not (very) suitable for help2man, thus I patched them a little bit.

1. Add short description in the beginning.
2. Move "Options:" and "Examples:" keywords to separate lines. (And fix exact wording where needed.)
3. Reformat options list (e.g. add doublespaces between option and its description).
4. Did some more trivial reformatting.

After this I've got not ugly looking manpages for most binaries.[3]

Since I do not know project's requirements for build scripts and build-dependencies, I did not add manpage's generation and installation to GNUmakefile.am but to Debian build scripts.

[1] https://www.gnu.org/software/help2man/
[2] https://salsa.debian.org/debian/can-utils/blob/master/debian/rules
[3] https://manpages.debian.org/unstable/can-utils/


Mike351c:
Might need a linker, join, thread in c.
Put a general output file for each and edit the output once successful.
Python Java sub might work better pending your preference.
My 2 cents.
Mike

hartkopp:
Don't know what @Mike351c Mike351c want's to tell me, sorry ...

IMO the cleanup is very valuable to get man pages - or at least an excellent starting point for manual enhancements. I'll try that out and give some more feedback then.
Thanks @gerasiov

hartkopp:
Hello Alexander @gerasiov ,
I fixed some warnings and added the CAN FD support for cangw which will emerge in Linux 5.4.
While thinking about contacting you for a Debian can-utils update I found this pull request, which I unfortunately did not review to the end.
I would like to take this PR but now 4 files have been changed and therefore generate merge conflicts.
Would it be ok for you to rework the 4 patches based on the latest HEAD?
Best,
Oliver

gerasiov:
Don't worry, I'll review this patchset and update PR while package new version for Debian.

hartkopp:
Thanks!!

@marckleinebudde , @yegorich : Do you have anything that should be done before the Debian can-utils package from Feb 2018 https://tracker.debian.org/pkg/can-utils get's an update?

yegorich:
Is this a valid bug? https://stackoverflow.com/questions/57630731/candump-utility-not-printing-leading-zeros-for-data-length-on-non-fd-frames

marckleinebudde:
There are some j1939 related changes pending, @olerem has some patches pending.

hartkopp:
No this is intentional to distinguish Classic CAN [x] from CAN FD [xx] frames in the human readable output. The questioner is wrong by stating the DLC to be the same, as we always operate with the real length information. And when both are 5 bytes you wont be able to distinguish them.

hartkopp:
> There are some j1939 related changes pending, @olerem has some patches pending.

Ok. Then lets wait for @olerem 's j1939 patches and @gerasiov 's help2man patches.

I did not want to create any pressure. Just felt the Debian package could get an update also because of the SPDX cleanup from @yegorich .

hartkopp:
> Is this a valid bug? https://stackoverflow.com/questions/57630731/candump-utility-not-printing-leading-zeros-for-data-length-on-non-fd-frames

Just answered that question on Stack Overflow ... thanks for the hint.

marckleinebudde:
> > There are some j1939 related changes pending, @olerem has some patches pending.
> 
> Ok. Then lets wait for @olerem 's j1939 patches and @gerasiov 's help2man patches.
> 
> I did not want to create any pressure. Just felt the Debian package could get an update also because of the SPDX cleanup from @yegorich .

The j1939 patches by @olerem are merged.

yegorich:
I've used [reuse](https://reuse.software/tutorial/) to check for OSS "compliance" and got a lot of warnings. By the way, do we want to get rid of license texts in the source files?

gerasiov:
@yegorich
JFYI In Debian we use licensecheck and license-reconcile to generate/check debian/copyright file with actual copyright and license info on per-file basis.

Issue #75, opened 2018-03-24
cansend. Valgrind reports points to uninitialised byte(s) in bind call.


Shmink:
```
==7553== Syscall param socketcall.bind(my_addr.sa_data) points to uninitialised byte(s)
==7553==    at 0x4F420F7: bind (syscall-template.S:84)
==7553==    by 0x400CBA: ??? (in /usr/bin/cansend)
==7553==    by 0x4E5A82F: (below main) (libc-start.c:291)
==7553==  Address 0xffefffd12 is on thread 1's stack
```

The program still works fine but in my own project it was an issue and I fixed it by adding `bzero(&addr,sizeof(struct sockaddr_can));` to zero out the struct. I tried to fork and add it myself but after doing that then running `make` and the command `./cansend vcan0 123#DEADBEEF` I get the error message `write: No such device or address`. Still learning C so not to sure why it's the case but your repo has been a big help to me with my project.

hartkopp:
Hi Tom,

assuming vcan0 is up and running at that time: At which line number did you put the bzero() call?

Regards,
Oliver

Shmink:
`vcan0` is up and running. It would have been line 138 in `cansend.c`.

hartkopp:
You should move it in front of line 105 ... otherwise you kill the settings made there.

Shmink:
Yep, that sorted it and the Valgrind message has now gone. I'll submit a pull request.

PR #76, opened 2018-03-24
cansend. Valgrind was reporting uninitialised byte(s) in bind call.

Shmink:
Issue #75 for more details. Cansend still functions as before and Valgrind no longer picks up on this.

### DIFF
--- a/cansend.c
+++ b/cansend.c
@@ -102,6 +102,8 @@ int main(int argc, char **argv)
 		return 1;
 	}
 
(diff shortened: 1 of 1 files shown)


Shmink:
Very well.

Shmink:
Not in the latest commit that I can see. https://github.com/linux-can/can-utils/pull/76/commits/d8bbacfb476d5d26690cdb9910d15fb8600ea9ac

Shmink:
If you just want a single commit merge then I'll do it.

PR #93, opened 2018-08-22
cansend: send multiple frames in one invocation

arnout:
It is sometimes convenient to be able to send multiple frames with a single cansend invocation, otherwise the time between frames can get very large.

Split into two patches for easy review.

I added a 1ms delay between frames, otherwise it is too easy to overflow the transmit queue. Apparently write() is non-blocking, it returns -ENOBUFS. I suspect that adding a select() wouldn't work anyway. 1ms is used in other CAN utils as well, so I used that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ subsystem (aka SocketCAN):
 
 * candump : display, filter and log CAN data to files
 * canplayer : replay CAN logfiles
-* cansend : send a single frame
+* cansend : send single frames
 * cangen : generate (random) CAN traffic
 * cansniffer : display CAN data content differences (just 11bit CAN IDs)
 
--- a/cansend.c
+++ b/cansend.c
@@ -71,23 +71,6 @@ int main(int argc, char **argv)
 		return 1;
 	}
 
-	/* parse CAN frame */
-	required_mtu = parse_canframe(argv[2], &frame);
-	if (!required_mtu){
-		fprintf(stderr, "\nWrong CAN-frame format! Try:\n\n");
-		fprintf(stderr, "    <can_id>#{R|data}          for CAN 2.0 frames\n");
-		fprintf(stderr, "    <can_id>##<flags>{data}    for CAN FD frames\n\n");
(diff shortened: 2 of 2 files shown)


hartkopp:
I have not seen the announced 1ms delay in the code - which I think is good. The 1ms delay are pretty arbitrarily and I don't know if it would fit all use-cases. Where have you found this "1ms delay" in other can-utils applications?
Have you ever tried to use canplayer to send a sequence of time-wise separated CAN frames, like
 echo $'(0.000000) vcan0 234#112233\n(0.500000) vcan0 123#112233\n(0.800000) vcan0 456#112233' | canplayer
??

arnout:
I completely agree that it's arbitrary; I searched for another option but couldn't find any, except something very heavy-weight like using netlink to query the Tx queue length. In fact, for my application I have to change the 1ms in 4ms because the receiver is apparently unable to handle more than 2 frames back-to-back.

There are actually only two places in the code where 1ms sleep is used:

canfdtest.c-266-                        if ((counter % 33) == 0)
canfdtest.c:267:                                millisleep(3);
canfdtest.c-268-                        else
canfdtest.c:269:                                millisleep(1);

cangen.c-460-           if (gap) /* gap == 0 => performance test :-] */
cangen.c:461:                   if (nanosleep(&ts, NULL))
cangen.c-462-                           return 1;
(gap defaults to 1, is converted from milliseconds to ts)

cangen actually handles the ENOBUFS by doing a poll() and retrying.

The other 1ms sleeps are for bind() errors, not for write(). My bad.


Now that I think about it again, there is one other option for inserting a delay: use the loopback to detect when the frame has actually been sent. But that also makes the code a lot more complicated.


canplayer is indeed a much easier solution. Stupid that I didn't think of that. So feel free to close this PR.

hartkopp:
Thanks! I like these kind of requests as it points out the different use-cases you might want to perform with the can-utils. Maybe the echo command is not very handy - but I crafted different log-files by hand to fire a sequence of CAN frames in a specific order and timing for some easy hacks. At least I got my job done with canplayer and the crafted logfiles which can be stored in a repository too.

Best regards,
Oliver
(closing the PR)

PR #97, opened 2018-09-13
CI: treat gcc warnings as errors

yegorich:
For now exclude unused-result warnings in order not to
affect current builds.


yegorich:
The next step should be to add return value checks and make CI break when a return value isn't checked.

PR #99, opened 2018-09-18
Extend canfdtest to test proper RX and TX-ECHO order

olerem:


canfdtest.c
```diff
@@ -47,6 +47,7 @@ static int running = 1;
 static int verbose;
 static int sockfd;
 static int test_loops;
+static int exit_sig = 0;
```
marckleinebudde: don't initialzie statics as `0`

canfdtest.c
```diff
@@ -339,6 +352,10 @@ int main(int argc, char *argv[])
 		return 1;
 	}
 
+	if (echo_gen)
+		setsockopt(sockfd, SOL_CAN_RAW, CAN_RAW_RECV_OWN_MSGS,
```
marckleinebudde: error handling?

PR #100, opened 2018-09-18
Add missing return value checks

yegorich:


yegorich:
What is the preferred way of checking whether `write()` call failed? What if the function returns 0? For me it means that the call failed too. Or should we just ignore this case?

Another style question. Some utilities use `error()` calls to show error and/or exit instead of using `perror()`. Should this be made consistent?

marckleinebudde:
Feel free to make the error output consistent. Regarding `write()`, it returns the number bytes written, if it returns `0` something is wrong. For longer running programs better check the return value of write properly, it it's `-1` and `errno == EINTR` doens't mean it's an error, just a signal arrived.

PR #102, opened 2018-09-21
port canconfig from old canutils

olerem:
canconfig is still used, set lets move it here and completely
deprecate old tools.

Signed-off-by: Oleksij Rempel <o.rempel@pengutronix.de>


marckleinebudde:
What about merging libsocketcan into here first?

PR #109, opened 2018-10-10
Revert "Fix CAN ID typo in candump usage, to 12345678"

nick-ike:
Reverts linux-can/can-utils#108

The documentation was correct - the MSB in the CAN ID and mask must be set. My mistake. Thank you @hartkopp !

(I think this is the extended ID bit?)

### DIFF
--- a/candump.c
+++ b/candump.c
@@ -152,7 +152,7 @@ void print_usage(char *prg)
 	fprintf(stderr, "%s -c -c -ta can0,123:7FF,400:700,#000000FF can2,400~7F0 can3 can8\n", prg);
 	fprintf(stderr, "%s -l any,0~0,#FFFFFFFF    (log only error frames but no(!) data frames)\n", prg);
 	fprintf(stderr, "%s -l any,0:0,#FFFFFFFF    (log error frames and also all data frames)\n", prg);
-	fprintf(stderr, "%s vcan2,12345678:DFFFFFFF (match only for extended CAN ID 12345678)\n", prg);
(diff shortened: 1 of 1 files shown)


marckleinebudde:
What about improving the documentation and stating stat `0x80000000` is the extended CAN id bit?

hartkopp:
Good idea. I would put some sentence that the filter content is a raw value that is put into the kernel as-is.
IMO this makes more sense than checking the length of the CAN-ID to be 8 characters and then set the CAN_EFF_FLAG implicit ...

marckleinebudde:
Makes sense

nick-ike:
I haven't been able to completely unconfuse myself yet. 

I think that (and correct if wrong)
0x8000 0000 is CAN_EFF_FLAG as used by SocketCAN
SocketCAN automatically manages the CAN "Identifier extension bit (IDE)" which follows the first 11 bits of ID. (https://en.wikipedia.org/wiki/CAN_bus#ID_allocation)

The candump usage states, "When can_id and can_mask are both 8 digits, they are assumed to be 29 bit EFF.", that line is a bit confusing given that CAN_EFF_FLAG must be ORed in manually.

Thoughts on the commit with added CAN_EFF_FLAG documentation? Can change anything.

hartkopp:
Hi Nick,
the documentation had indeed another issue as you pointed out in your text above.
I added the 8 digit detection and clarified the documentation.
Finally candump now behaves as you originally expected:
candump vcan2,12345678:DFFFFFFF now filters like candump vcan2,92345678:DFFFFFFF

Thanks,
Oliver

ps. Added your credentials by "Reported-by:  Nick ..." in the commit.

PR #118, opened 2019-01-10
Fix unaligned pointer access.

twasilczyk:
This is an issue for Android (ARM) platform, throws a compiler warning.


cangw.c
```diff
@@ -330,9 +331,10 @@ int parse_mod(char *optarg, struct modattr *modmsg)
 		ptr++;
 	}
 
-	if (sscanf(++ptr, "%x.%hhx.%16s", &modmsg->cf.can_id,
```
hartkopp: Does the warning disappear when you cast the value with
(canid_t *)&modmsg->cf.can_id
??

twasilczyk: No, it doesn't.

/usr/local/google/home/twasilczyk/android/adler/can-utils/cangw.c:333:48: warning: taking address of packed member 'cf' of class or structure 'modattr' may result in an unaligned pointer value [-Waddress-of-packed-member]
        if (sscanf(++ptr, "%x.%hhx.%16s", (canid_t *)&modmsg->cf.can_id,
                                                      ^~~~~~~~~~~~~~~~~
1 warning generated.

This is not just a warning - ARM cpus may crash on unaligned pointer access.

hartkopp: I googled the warning - and clang 4.x just creates a false positive here. The can_id is the first and aligned value in the struct modattr. So no chance to make ARM CPUs crash on this.
Even if it doesn't win a design price I would apply your patch. But the commit message should not contain a "fix unaligned pointer access" (which is simply wrong) but something like "silence false positive clang warning" where you should point out that the warning does not fit to the data structure in question.

twasilczyk: Good point, I'm just trying to debug an issue on my system and fixing everything that's fishy.

hartkopp: Merged you patch. For now it looks the best approach to go.
Happy debugging!
BR Oliver

PR #120, opened 2019-01-10
Check if can_update_spt succeeded

twasilczyk:
This fixes uninitialized value compiler warning for tseg1 and tseg2.


marckleinebudde:
This is a copy of the kernel code, please fix the kernel code first.

PR #123, opened 2019-01-15
Don't fail when it wasn't possible to fetch the interface name.

twasilczyk:
On some systems `SIOCGIFNAME` may fail with `ENOTTY`, but the actual
`slcanX` interface gets properly configured. Instead of crashing hard
on such case, let's gracefuly degrade and just not display the
interface name.

This is the issue I mentioned in https://github.com/linux-can/can-utils/pull/118

### DIFF
--- a/slcand.c
+++ b/slcand.c
@@ -370,8 +370,13 @@ int main(int argc, char *argv[])
 	
 	/* retrieve the name of the created CAN netdevice */
 	if (ioctl(fd, SIOCGIFNAME, ifr.ifr_name) < 0) {
-		perror("ioctl SIOCGIFNAME");
-		exit(EXIT_FAILURE);
+		if (name) {
+			perror("ioctl SIOCGIFNAME");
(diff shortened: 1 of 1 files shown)


marckleinebudde:
On which system do you see the `ioctl()` to fail?

twasilczyk:
It's Android. I've built the kernel (4.4) by myself, so there is a chance I misconfigured something.

It's clearly a problem with my system, but it's hard to say if it's tied to Android P or my specific device. I'd appreciate help debugging this, but the patch with workaround has still some value, because any other Android developer may hit the issue at other time.

twasilczyk:
Bump, would you be willing to accept this PR to not hard fail there? It was not just "my" device, but indeed a target that's not very popular for CAN bus (Android/AOSP).

Either way, if I look into this PR again in 6 years, I may no longer work at Google and this won't be my concern anymore :).

PR #137, opened 2019-05-09
Errqueue

marckleinebudde:
Picking up @olerem work from https://github.com/linux-can/can-utils/pull/135.


marckleinebudde:
Just for testing

PR #138, opened 2019-05-09
Errqueue broken

marckleinebudde:
Picking up @olerem work from https://github.com/linux-can/can-utils/pull/135.


marckleinebudde:
This was just for testing.

PR #139, opened 2019-05-28
Add uds output functionality

buecheko:
Adds the option -u to isotpdump which will print the UDS message type as well as the service name or response code.

### DIFF
--- a/isotpdump.c
+++ b/isotpdump.c
@@ -47,6 +47,7 @@
 #include <string.h>
 #include <libgen.h>
 #include <time.h>
+#include <strings.h>
 
 #include <net/if.h>
 #include <sys/types.h>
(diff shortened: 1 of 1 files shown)


isotpdump.c
```diff
@@ -362,6 +377,115 @@ int main(int argc, char **argv)
...
+				if (uds_output && is_ff) {
+					int service = frame.data[datidx];
+					char * service_name;
+					int offset = 3;
+					char flag[] = "[???]";
+					if (asc)
+						offset = 1;
+					printf("%*s", ((7-ext) - (frame.len-datidx))*offset + 3,
+					       " - ");
+					if ((service >= 0x50 && service <= 0x7E) || (service >= 0xC3 && service <= 0xC8)) {
```
hartkopp: Good idea.
Can you put the following into a function like print_uds_output(int service, char *service_name, char *flag) ?
The indention grows too much and you may omit the service variable here.

buecheko: Good point. I moved most of it to a function like you mentioned. I am not 100% sure what you meant with "you may omit the service variable". I hope it's okay that way.

hartkopp:
Yes, the updated patch was exactly my intention.
Thanks for the contribution,
Oliver

PR #142, opened 2019-06-13
j1939 : Allows to receive own frame.

arthurguyader:
Need this patch after commit on kernel.org 	:
dfa62204a6f5c8689431a21b23830a5929b373d1 j1939: socket: rework receive filter.

Signed-off-by: Arthur Guyader <arthur.guyader@iot.bzh>


marckleinebudde:
In the mean time, we've removed `SO_J1939_RECV_OWN` from the kernel, as it doesn't work for ETP frames.

PR #145, opened 2019-07-09
add j1939 tests

olerem:
Signed-off-by: Oleksij Rempel <o.rempel@pengutronix.de>


hartkopp:
Hello Oleksij,
when this are tests for the CAN subsystem / CAN protocol implementations like CAN_RAW, CAN_BCM or CAN_J1939 we should use https://github.com/linux-can/can-tests for that.

marckleinebudde:
I'm not sure if we should clutter the user space tools over several repos....

hartkopp:
Of course not. But userspace tools are tools for the *user* to use SocketCAN.
When you want to test the in-kernel implementation and corner cases like here:
https://github.com/linux-can/can-tests/blob/master/test_netlayer.sh
or https://github.com/linux-can/can-tests/blob/master/tst-bcm-cycle.c
This should not blow up the user space tools for people that do/should not care about kernel internals.

marckleinebudde:
Ok, the test will not be pushed into the can-utils repo.

hartkopp:
Thanks. I thought about it over night and I would propose to bring some more structure into can-tests. E.g. create some directories like bcm, raw, gw, j1939, netlayer and put the specific test and examples right there.
Having a place for testing and sample code is valuable IMO. But we need to clean that up to have a handy and easy to use environment to point people to.

marckleinebudde:
Make it so!

hartkopp:
We reworked the https://github.com/linux-can/can-tests repository for some time now.
@olerem : Would you like to prepare the content to be integrated there?

marckleinebudde:
The j1939 test have been added to the can-test repo:
https://github.com/linux-can/can-tests/pull/7

PR #155, opened 2019-09-20
Reuse fixes

yegorich:
This patch series fixes some of the issues mentioned by the REUSE utility

### DIFF
--- a/LICENSES/GPL-2.0-only.txt
+++ b/LICENSES/GPL-2.0-only.txt
@@ -1,0 +1,340 @@
+		    GNU GENERAL PUBLIC LICENSE
+		       Version 2, June 1991
+
+ Copyright (C) 1989, 1991 Free Software Foundation, Inc.
+     59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+			    Preamble
+
+  The licenses for most software are designed to take away your
+freedom to share and change it.  By contrast, the GNU General Public
+License is intended to guarantee your freedom to share and change free
+software--to make sure the software is free for all its users.	This
+General Public License applies to most of the Free Software
+Foundation's software and to any other program whose authors commit to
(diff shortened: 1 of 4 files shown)


LICENSES/GPL-2.0-only.txt
```diff
@@ -0,0 +1,340 @@
...
+REDISTRIBUTE THE PROGRAM AS PERMITTED ABOVE, BE LIABLE TO YOU FOR DAMAGES,
+INCLUDING ANY GENERAL, SPECIAL, INCIDENTAL OR CONSEQUENTIAL DAMAGES ARISING
+OUT OF THE USE OR INABILITY TO USE THE PROGRAM (INCLUDING BUT NOT LIMITED
+TO LOSS OF DATA OR DATA BEING RENDERED INACCURATE OR LOSSES SUSTAINED BY
+YOU OR THIRD PARTIES OR A FAILURE OF THE PROGRAM TO OPERATE WITH ANY OTHER
+PROGRAMS), EVEN IF SUCH HOLDER OR OTHER PARTY HAS BEEN ADVISED OF THE
+POSSIBILITY OF SUCH DAMAGES.
+
+		     END OF TERMS AND CONDITIONS
+
```
hartkopp: I was not aware of the following how-to.
Are you sure we would need to add that following text and not end the license file here?

yegorich: This is the end the license file IMHO. This approach is suggested by REUSE project and Linux kernel also follows this structure.

hartkopp: Ok. I wasn't aware that this how-to is really part of the license text.
Strange but hey ... then it's obviously the right thing to do.
Acked-by: Oliver Hartkopp <socketcan@hartkopp.net>

yegorich:
There is one more questio left. Linux kernel provides license files with additional fields (https://github.com/torvalds/linux/blob/master/LICENSES/preferred/BSD-3-Clause). So should we this approach or just put the license text only?

yegorich:
v2: added Oliver's Ack and replaced BSD license with text only version as the other two.

PR #167, opened 2019-11-12
libj1939: fix missing include and syntax

yegorich:
Include err.h and remove the unneeded parameter errno and
its header include.

Fixes: 791b6de ("treewide: Fix compilation on libmusl, replace error() by err()")


marckleinebudde:
Fixes: 791b6de78673 ("treewide: Fix compilation on libmusl, replace error() by err()")

PR #169, opened 2019-11-13
Isotpsniffer add llopts support

rumpelsepp:


isotpsniffer.c
```diff
@@ -302,6 +315,17 @@ int main(int argc, char **argv)
 
 	setsockopt(t, SOL_CAN_ISOTP, CAN_ISOTP_OPTS, &opts, sizeof(opts));
 
+	if ((setsockopt(s, SOL_CAN_ISOTP, CAN_ISOTP_LL_OPTS, &llopts, sizeof(llopts))) < 0) {
+		perror("setsockopt");
+		close(s);
```
hartkopp: is close(s) needed before exit?
AFAIK exit closes all file descriptors

isotpsniffer.c
```diff
@@ -302,6 +315,17 @@ int main(int argc, char **argv)
...
+		exit(1);
+	}
+	if ((setsockopt(t, SOL_CAN_ISOTP, CAN_ISOTP_LL_OPTS, &llopts, sizeof(llopts))) < 0) {
+		perror("setsockopt");
+		close(s);
```
hartkopp: dito

isotpsniffer.c
rumpelsepp: It does, but I learned in other projects that it is "good style" to close them explicitely.

rumpelsepp:
> The if statement checks for "printable" ASCII chars - not for ASCII chars that are used in hex numbers. You were probably thinking of isprint() ?!?

You are right, I committed  a faulty version, since I did some experimentations.

isotpsniffer.c
hartkopp: But then both sockets (s & t) need to be taken care of, right?

marckleinebudde: If you want to have proper error handling, please add it in a separate patch. Please use goto-style error handling, if you have more than one resource.

rumpelsepp: yep. better to fix the error handling properly before merging. will update this PR.

PR #175, opened 2019-11-29
sync testj1939 with latest kernel UAPI

olerem:


can-j1939-kickstart.md
```diff
@@ -59,15 +59,15 @@ is not meant for us and *testj1939* does not receive it.
 Binding a can-j1939 socket to a source address will register
 allow you to send packets.
```
josemic: I do not understand the sentence. Please rephrase.

olerem: This is not related to this patch. Usually a patch should be limited to one change or one sort of changes. If described issues is not related to this patch, then it should be reported or patched in different commit.
If patch is reviewed, review should be related to the patch. If patch has missing changes or changes are wrong, then review can prevent passing it to go upstream.

josemic: Ok. Makes sense. Added bug #178 for that issue.

PR #189, opened 2019-12-23
candump: splitting log files by size

BugLight:
We are using candump to log CAN traffic on standalone devices. The log file grows fast. We suggest log file splitting. It could be helpful in some cases like in our one.

### DIFF
--- a/candump.c
+++ b/candump.c
@@ -114,6 +114,9 @@ extern int optind, opterr, optopt;
 
 static volatile int running = 1;
 
+static int part = 0;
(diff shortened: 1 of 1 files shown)


hartkopp:
Generally a good idea - but it bloats candump once more. E.g. I'm currently thinking of removing options (like -b and -B) to meet Unix tool simplicity philosophy ("Write programs that do one thing and do it well.")
Have you ever thought about `candump -L can0 | split -C 500MB`?
Are you familiar with `split`?

PR #191, opened 2020-01-04
candump: remove bridging functionality

hartkopp:
The tool 'candump' is used to dump CAN traffic to stdout or logfile.
In early days where the in-kernel CAN gateway (can-gw) with the cangw tool was
not yet implemented the bridging functionality allowed to send received (and
filtered) CAN frames to another outgoing CAN interface.

As we now have can-gw and sending CAN frames from a 'reading tool' seems wrong
especially from the Unix philosophy "Write programs that do one thing and do
it well." this patch removes the obsolete functionality from candump.

Signed-off-by: Oliver Hartkopp <socketcan@hartkopp.net>


yegorich:
@hartkopp have never used this feature. Is it still working?

hartkopp:
This is exactly my expectation - that nobody really knew about this feature and nobody ever used it.
We used this feature in some prototype until we had so many routings that it just failed with the latencies and the scheduling. That's why I created can-gw ...

hartkopp:
Thanks for your feedback!
Thee I will merge it right now - live from the "Software Defined Networking devroom" at FOSDEM 2020 in Bruxelles :-)
https://fosdem.org/2020/schedule/track/software_defined_networking/

PR #195, opened 2020-02-14
Fix compiler warnings in cangen

a3f:
Nothing serious, but I got two warnings when opening the file with my clang (YouCompleteMe) setup.

### DIFF
--- a/cangen.c
+++ b/cangen.c
@@ -53,6 +53,7 @@
 #include <libgen.h>
 #include <time.h>
 #include <errno.h>
+#include <getopt.h>
 
 #include <sys/time.h>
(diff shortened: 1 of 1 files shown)


a3f:
> Can you tell me why this showed up in cangen.c but not in candump.c then?

It's a linter running on files I open. I haven't took a look at the other files.

I've educated myself some more and getopt indeed comes out of `<unistd.h>` given suitable feature macros, which I didn't define. Making functions static that aren't usable outside is still applicable, but if one wants to do it, all files should probably be touched.

Please dismiss. Sorry for the noise.

hartkopp:
Ok. Thanks!

PR #204, opened 2020-05-21
Change path for canutils modules

PsycheEYE:
Change path for canutils modules from system/bin to vendor/bin in case of building in AOSP tree


gibsson:
That would indeed be a good thing to have.
However have you made sure that option doesn't break the build on older AOSP version (where LOCAL_VENDOR_MODULE doesn't exist)?

PsycheEYE:
AFIAK previous versions of Android build system simply ignore that variable(LOCAL_VENDOR_MODULE), but I'll check it.
Also i'll try to add j1939 stuff building logic

hartkopp:
Please change the commit description / message to something like:

Android.mk: Introduce LOCAL_VENDOR_MODULE to build in AOSP tree

Change path for canutils modules from system/bin to vendor/bin in case of building in AOSP tree.

hartkopp:
@PsycheEYE @gibsson 
I updated both the target list and the module path here:
https://github.com/linux-can/can-utils/pull/215
Given that the Travis build was fine - can please check whether my branch at https://github.com/hartkopp/can-utils/tree/Android.mk-Updates really builds the desired targets on your system?
Many thanks for your support!
Oliver

PR #213, opened 2020-05-29
Added CAN message direction (Tx or Rx) for log recording

brpallu:
Log file recording updated with CAN Message direction.
Changes Made:
    1.Added CAN message direction (Tx or Rx) for log file recording in candump.
    2.Log file converters log2asc and asc2log are updated to parse the message direction.

NOTE:
Log2asc was appending "Rx" for all the recorded can messages for conversion. With this the trace was not much helpful for analysis. So they are updated to reflect correct correct message direction.

### DIFF
--- a/asc2log.c
+++ b/asc2log.c
@@ -72,6 +72,8 @@ void print_usage(char *prg)
 
 void prframe(FILE *file, struct timeval *tv, int dev, struct canfd_frame *cf, unsigned int max_dlen, char *msgdir) {
 
+	char buf[CL_CFSZ]; /* max length */
+
 	fprintf(file, "(%ld.%06ld) ", tv->tv_sec, tv->tv_usec);
 
 	if (dev > 0)
@@ -79,9 +81,11 @@ void prframe(FILE *file, struct timeval *tv, int dev, struct canfd_frame *cf, un
 	else
 		fprintf(file, "canX ");
 
-	fprintf(file, "%s ", msgdir);
+	sprint_canframe(buf, cf, 0, max_dlen);
+	fprintf(file, "%-25s ", buf);
 
-	fprint_canframe(file, cf, "\n", 0, max_dlen);
+	fprintf(file, "%s \n", msgdir);
+
 }
 
 void get_can_id(struct canfd_frame *cf, char *idstring, int base) {
--- a/asc2log.c
+++ b/asc2log.c
@@ -82,7 +82,7 @@ void prframe(FILE *file, struct timeval *tv, int dev, struct canfd_frame *cf, un
 		fprintf(file, "canX ");
 
 	sprint_canframe(buf, cf, 0, max_dlen);
-	fprintf(file, "%-25s ", buf);
+	fprintf(file, "%s ", buf);
 
 	fprintf(file, "%s \n", msgdir);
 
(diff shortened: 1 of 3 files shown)


hartkopp:
The candump tool logs a CAN bus which is a multi-master bus where you can not distinguish the 'sender' of a frame. As many other details that are logged in the ASC file format (see recent changes for CAN FD support in asc2log), the direction information is pretty pointless in real world use-cases.
And the weird ASC file format specification is definitely not the benchmark for a good ASCII CAN bus logfile design. Therefore I do not tend to add any useless complexity to Linux tools to increase the compatibility for the today seldom used ASC file format from a proprietary vendor.

brpallu:
Yes, the CAN bus is multi-master and you cannot distinguish the sender. But you can always know if the message was sent or received by the host.
I don't agree with the comment that the direction is pointless as one may need to know what are the can messages received and sent out. If it was the case the candump standard output with option '-x' should never have included TX & RX information.

This logs2asc file conversion tool converts candump log to ASC format with message direction as Rx hard coded. I think its better and useful if we have this updated with correct value instead of the hard coded one.

hartkopp:
The problem is, that even if I think very benevolent about adding the rx/tx information in the log file format, your approach breaks every CAN logfile parser since 2006.
There was an extension made for CAN FD where the legacy tools (intentionally) failed when a CAN FD frame appeared in the logfile. When adding new content behind the CAN interface name all existing logfiles become unusable!
Therefore an extension might probably happen in the form of a newly introduced separate string at the end of the logfile line - which **has to be ignored by existing canplayer implementations**.
The rx/tx information is not relevant to canplayer anyway when canplayer is sending the content from a logfile.

hartkopp:
Oh, I programmed that approach myself, sorry.
See https://github.com/linux-can/can-utils/tree/logfile-flags

brpallu:
Yes I completely agree. Thank you for your suggestion. Appending the Tx/Rx at the end of can log file will facilitate the new tools to be backward compatible with old log files. Also canplayer will not break and continue to work as expected.

hartkopp:
I think we should give this extension some testing before adding it to the master branch.
Maybe you or someone else will find an issue at daily work.
So far I don't see an issue with this approach.
Thanks,
Oliver

brpallu:
I did try testing my changes with old and new set of can logs. Both log2asc and asc2log look  pretty good in conversion. Also canplayer can play old can log, new can log and also converted can log files without issues.

hartkopp:
Thanks for testing!
I will post a RFC about the change in https://github.com/linux-can/can-utils/tree/logfile-flags on the linux-can mailing list, before merging it to the master branch.

brpallu:
Hi Oliver,
I validated only my changes and haven't tested yours. And also I see there is an additional flag passed as part of candump log recording in your commit which increases the complexity for using the tool.
The additional flag is unnecessary and whereas changes in this PR is backward compatible without the need of additional flag which makes the tool much simpler to use.
I feel this PR should be merged as the changes have been already validated.

Regards,
Pallavi

hartkopp:
Hello Pallavi,

your PR has several style issues and I wanted to point out in my implementation that the extra information can be extended for future use and does not only stick to the message direction.

Of course I attributed that it was _your_ suggestion in the commit message!

About the '-x' flag: Nobody asked for the direction information or other stuff for about 14(!) years now.
And therefore I don't want to bloat the logfile size by default. The extra_flags (maybe I should rename it to extra_info) remain optional to the users that need this functionality.

Regards,
Oliver

PR #214, opened 2020-05-29
J1939 clang fixes

hartkopp:
Fix clang warnings in recently added j1939 files.
See Garys fixes here:
https://github.com/linux-can/can-utils/commit/43610bd6217cd8b38455ea9af6783fc04547929f
and here:
https://github.com/linux-can/can-utils/commit/46895a41c5241971f48bf936056de597d8190e29

### DIFF
--- a/j1939cat.c
+++ b/j1939cat.c
@@ -165,11 +165,11 @@ static void j1939cat_scm_opt_stats(struct j1939cat_priv *priv, void *buf, int le
 	int offset = 0;
 
 	while (offset < len) {
-		struct nlattr *nla = (struct nlattr *) (buf + offset);
+		struct nlattr *nla = (struct nlattr *) ((char *)buf + offset);
 
 		switch (nla->nla_type) {
 		case J1939_NLA_BYTES_ACKED:
-			stats->send = *(uint32_t *)((void *)nla + NLA_HDRLEN);
+			stats->send = *(uint32_t *)((char *)nla + NLA_HDRLEN);
 			break;
 		default:
 			warnx("not supported J1939_NLA field\n");
--- a/j1939cat.c
+++ b/j1939cat.c
@@ -121,7 +121,7 @@ static ssize_t j1939cat_send_one(struct j1939cat_priv *priv, int out_fd,
 		return -EINVAL;
 	}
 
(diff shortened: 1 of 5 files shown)


testj1939.c
```diff
@@ -105,7 +105,7 @@ int main(int argc, char *argv[])
 		break;
 	case 's':
 		todo_send = strtoul(optarg ?: "8", NULL, 0);
-		if (todo_send > sizeof(dat))
+		if (todo_send > (int)sizeof(dat))
```
olerem, line 108: Hm... strtoul is unsigned long int:
unsigned long int strtoul(const char *nptr, char **endptr, int base);
The sizeof should be size_t, which is unsigned int or unsigned long int, depending on the system.

Why it should be int?

j1939cat.c
```diff
@@ -121,7 +121,7 @@ static ssize_t j1939cat_send_one(struct j1939cat_priv *priv, int out_fd,
 		return -EINVAL;
 	}
 
-	if (num_sent > buf_size) /* Should never happen */ {
+	if ((size_t)num_sent > buf_size) /* Should never happen */ {
```
olerem, line 124: Here is better to change type of buf_size to ssize_t. We use ssize_t by calling this function any way.

testj1939.c
```diff
@@ -231,7 +231,7 @@ int main(int argc, char *argv[])
 
 	if (todo_send) {
 		/* initialize test vector */
-		for (j = 0; j < sizeof(dat); ++j)
+		for (j = 0; j < (int)sizeof(dat); ++j)
```
olerem, line 234: "j" can be changed to unsigned int

testj1939.c
hartkopp: todo_send is defined as int.
Is it save to define todo_send as unsigned int ?
sizeof(dat) should never be as high as ULONG_MAX but I was confused about the potential return values given in the man page.

olerem: If i see it correctly, we will get always unsigned value. On error, we will get ULONG_MAX, in case of strtoul().

hartkopp: Ok, then I'll make todo_send an unsigned int.
Thanks!

PR #221, opened 2020-06-17
Changed hardcoded hex number 0x0fed8 to j1939.h constant "J1939_PGN_ADDRESS_COMMANDED"

josemic:
Use include constant instead of hardcoded constant


josemic:
In line 634 one aditional hardcoded hex number has to be changed to the constant J1939_PGN_ADDRESS_COMMANDED

marckleinebudde:
> In line 634 one aditional hardcoded hex number has to be changed to the constant J1939_PGN_ADDRESS_COMMANDED

See https://github.com/linux-can/can-utils/pull/223

PR #239, opened 2020-09-23
added script to run on power

genisysram:
Thanks for the code, its working good.
## What do these changes do?
Added Architecture "ppc64le"
## Are there changes in behavior for the user?
No

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ matrix:
           before_script:
               - mkdir gcc_rel && cd gcc_rel
               - cmake ..
-          script:  make
+          script: make
 
         - env: test="x64 5.0 (CMake clang)"
           os: linux
@@ -36,7 +36,7 @@ matrix:
           before_script:
               - mkdir clang_rel && cd clang_rel
               - cmake ..
-          script:  make
+          script: make
 
         - env: test="x64 5.0 (CMake clang - Ninja) Debug"
           os: linux
@@ -53,7 +53,7 @@ matrix:
           android:
             components:
               - build-tools-25.0.2
-              #Including script to run on power
+ #Including script to run on power
         - env: test="x64 4.8.4 (make)"
           os: linux
           dist: trusty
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -53,6 +53,57 @@ matrix:
           android:
             components:
               - build-tools-25.0.2
+              #Including script to run on power
```
marckleinebudde, line 56: The indention looks wrong, better start the `#` at the beginning of a new line.

.travis.yml
```diff
@@ -53,6 +53,57 @@ matrix:
...
+          arch: ppc64le
+
+        - env: test="x64 4.8.4 (CMake gcc)"
+          os: linux
+          dist: trusty
+          compiler: gcc
+          before_script:
+              - mkdir gcc_rel && cd gcc_rel
+              - cmake ..
+          script:  make
```
marckleinebudde, line 85: Please just on space before `make`

.travis.yml
```diff
@@ -53,6 +53,57 @@ matrix:
...
+          arch: ppc64le
+
+        - env: test="x64 5.0 (CMake clang)"
+          os: linux
+          dist: trusty
+          compiler: clang
+          before_script:
+              - mkdir clang_rel && cd clang_rel
+              - cmake ..
+          script:  make
```
marckleinebudde, line 95: same here

marckleinebudde (changes requested):
It looks you added the new test targets in the middle of the `Android NDK (CMake)`. I think Line 49 is a good way to add new ones.

.travis.yml
```diff
@@ -53,7 +53,7 @@ matrix:
           android:
             components:
               - build-tools-25.0.2
-              #Including script to run on power
+ #Including script to run on power
```
marckleinebudde, line 56: You add the new test targets in the middle of the Android NDK (CMake). I think Line 49 is a good way to add new ones.

marckleinebudde (changes requested):
Please fix the location where you add the new tests.

marckleinebudde:
Hmm, look at the diff against `origin/master`, that doesn't look good.

marckleinebudde:
Have a look again at your first patch, you add the new tests in the middle of the android test, this is why it breaks. I've fixed the patch and created a new pull request: https://github.com/linux-can/can-utils/pull/242

PR #249, opened 2020-10-12
convert usleep to nanosleep

neheb:
usleep is removed in POSIX 2008.

### DIFF
--- a/bcmserver.c
+++ b/bcmserver.c
@@ -114,6 +114,7 @@
 #include <string.h>
 #include <signal.h>
 #include <errno.h>
+#include <time.h>
 
 #include <sys/types.h>
 #include <sys/wait.h>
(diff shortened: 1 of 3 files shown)


bcmserver.c
```diff
@@ -179,9 +180,11 @@ int main(void)
 	saddr.sin_addr.s_addr = htonl(INADDR_ANY);
 	saddr.sin_port = htons(PORT);
 
+	struct timespec r;
 	while(bind(sl,(struct sockaddr*)&saddr, sizeof(saddr)) < 0) {
 		printf(".");fflush(NULL);
-		usleep(100000);
+		struct timespec f = {0, 100 * 1000 * 1000};
```
marckleinebudde, line 186: Please move the declaration of the struct to the beginning of the scope, please initialize it using C99 initializers.

```
struct timesepc f = {
	.tv_nsec = 100 * 1000 * 1000,
};
```

bcmserver.c
```diff
@@ -179,9 +180,11 @@ int main(void)
...
+		nanosleep(&f, &r);
```
marckleinebudde, line 187: You're not using `&r`, so please use `NULL` intead.

marckleinebudde (changes requested):
The comments from the first file applies to the others, too.

bcmserver.c
neheb: Done.

bcmserver.c
neheb: I added r as this is in a loop.

marckleinebudde: But `r` is not used, why do you need it?

neheb: Fixed.

marckleinebudde (approved):

PR #252, opened 2020-10-13
change {} to {0}

neheb:
The former throws a pedantic warning.


marckleinebudde (changes requested):
Looks good, one nitpick. Can you add spaces around the 0? So that it reads `= { 0 };`

neheb:
meh.

hartkopp:
ok - and the subject should be something more informative and not like "increase value by one"

e.g. "make struct initialization C99 compatible again"

marckleinebudde (approved):

hartkopp:
I changed the subject as suggested.
Thanks Rosen.

PR #258, opened 2020-11-04
Android.mk: remove duplicated cansequence definition

gibsson:
Otherwise throwing the following error:
build/make/core/base_rules.mk:325: error: external/can-utils:
MODULE.TARGET.EXECUTABLES.cansequence already defined by
external/can-utils.

Fixes: a726c2a ("Merge branch 'cansequence' of
github.com:marckleinebudde/can-utils into master")

Signed-off-by: Gary Bisson <gary.bisson@boundarydevices.com>

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -161,21 +161,6 @@ LOCAL_CFLAGS := $(PRIVATE_LOCAL_CFLAGS)
 include $(BUILD_EXECUTABLE)
 
 #
-# cansequence
-#
-
-include $(CLEAR_VARS)
-
-LOCAL_SRC_FILES := cansequence.c
-LOCAL_MODULE := cansequence
-LOCAL_MODULE_TAGS := optional
-LOCAL_STATIC_LIBRARIES := libcan
-LOCAL_C_INCLUDES := $(LOCAL_PATH)/include/
-LOCAL_CFLAGS := $(PRIVATE_LOCAL_CFLAGS)
-
-include $(BUILD_EXECUTABLE)
-
-#
 # bcmserver
 #
 


marckleinebudde:
Why didn't travis find this error?

gibsson:
I doubt Travis is building can-utils within Android source code?

marckleinebudde:
See travis config:
https://github.com/linux-can/can-utils/blob/master/.travis.yml#L121

Is there anything we can improve?

gibsson:
This checks that can-utils can build with NDK. It doesn't check that it builds within the Android tree. Unfortunately those 2 are different approaches.
While checking NDK build is a good thing, as it will check libc compatibility for instance, AOSP build uses soong/ninja which is different:
https://source.android.com/setup/build
You could get the Android source and use a generic target (like `aosp_arm-eng`) to build the repo, but that might be overkill considering how big the Android repo is.

marckleinebudde:
Have you any experience with AOSP automation in travis? How long would it take to download the Android repo and build the can-utils against it?

gibsson:
No I don't have much experience with Travis at all.
As for the time it takes to download the source, well it depends on your connection, because the overall source tree weighs about 250GB according to Google:
https://source.android.com/setup/build/requirements
Then building only can-utils should be fairly quick. But really here, since it builds with NDK, all you would need is for the Makefile to be parsed by Soong to check for such duplicate, there must be an easier way than pulling all the repos.

marckleinebudde:
If you know the commands how to set this up, we can integrate that into the travis config.

gibsson:
I actually looked for it and couldn't find a way to properly do it yet.
The build error came from this makefile FYI that checks that a LOCAL_MODULE is only defined once:
https://android.googlesource.com/platform/build/+/refs/heads/master/core/base_rules.mk#325
Maybe the easiest would be to create a custom script that checks for that issue...

PR #260, opened 2020-11-11
can-utils: support len8_dlc for Classical CAN frames

hartkopp:
This is the patch set that has been developed to support DLC values from 9 .. 15 for Classical CAN capable CAN adapters.


hartkopp:
Needs some rework and squashing some commits together after the final includes have been merged to net-next.

PR #266, opened 2020-12-04
Enable all timestamp options to be used in log format files and stdout

### LINKED ISSUE #265: candump support all timestamp formats in log files, and log format stdout
As a user, I want the option to log zero based relative timestamp formats to a file.  For instance, sometimes I don't care about absolute time, and relative is easier to read so I would like to log directly it directly to a file.  

**Issue Description**
canutils is currently unable to use timestamp type options in combination with log output -L or -l.
` -t <type>   (timestamp: (a)bsolute/(d)elta/(z)ero/(A)bsolute w date)`
and
`-l          (log CAN-frames into file.`
`-L          (use log file format on stdout)`

In regards to log output format compatibility.  Absolute, and zero logs formats work fine with can player.  So there is no reason they shouldn't be supported. 
Delta log format is different as it prints time between messages, so not suited for canplayer, unless it has a delta playback mode.  This is typically only used in specific diagnostic scenarios, I don't think we need to protect users from playing it back.

**Example usage**
```
candump -L -tz vcan0
(000.000000) vcan0 0667BAF6#89C76376D62ADD77
(000.200074) vcan0 05F541AB#
(000.400156) vcan0 187E0AA5#651E4230F28F
(000.600243) vcan0 0C091058#F45333551ABF4051

./candump -L -ta vcan0
(1607120998.528012) vcan0 0E7D3A28#0EE8651E732FBF64
(1607120998.728108) vcan0 00EDF658#48AE52014F

./candump -L -td vcan0
(000.000000) vcan0 1C7C20AE#77E569140F27AC41
(000.200083) vcan0 15E0F04D#EF
(000.200085) vcan0 008702CA#1B431B0CC811287B
```

**Fix Description**

Previously `-l if(log)` and `-L if(logformt)` bypassed all timestamp formatting by using `goto out_fflush`.  I have extracted the timestamp formatting to print_timestamp so timestamp options are now available to logging workflows.

**Known Issues**
canplayer does not support playing back -tA format logs
(2020-12-04 17:25:57.869400) vcan0 02E939C4#DE18475339

richnetdesign:
resolves #265

### DIFF
--- a/candump.c
+++ b/candump.c
@@ -87,6 +87,8 @@
 #define SILENT_ANI 1  /* silent mode with animation */
 #define SILENT_ON  2  /* silent mode (completely silent) */
 
+#define TIMESTAMPSZ 44 /* sprintf of timestamp format output between 14 and 44 bytes */
+
 #define BOLD    ATTBOLD
 #define RED     ATTBOLD FGRED
 #define GREEN   ATTBOLD FGGREEN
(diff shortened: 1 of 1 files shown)


hartkopp:
See comment at  https://github.com/linux-can/can-utils/issues/265

Issue #38, opened 2017-03-08
Replace the unportable signal with sigaction


fsasm:
According to the Linux manpage signal(2) the function `signal` should not be used as its behaviour can vary across platforms. Currently glibc mimics the BSD behaviour which implicitly uses the flag `SA_RESTART`. This behaviour can be different when other libc are used. Instead of `signal` `sigaction` should be used.

marckleinebudde:
Feel free to create a patch.

PR #264, opened 2020-12-03
candump: epoll_wait() instead of select() runs faster and better keeps packet order

IDC-Dragon:
Using epoll_wait() instead of select() gives higher performance for listening on multiple interfaces. Additionally, the read order has a higher chance to resemble the true temporal order. select() gives implicit priority to the lower index socket.

PR #268, opened 2020-12-05
Update signal handlers to use sigaction, and re-open logfile on SIGHUP to integrate logrotated support

richnetdesign:
1. Treats sighup as signal to close and reopen log file
2. Candump only supports autogenerated log file name by default.  For use with logrotation daemon it expects consistent filename.  I updated -l option to accept optional filename.
3. Since interface is a required argument, getopt skips final arg.  Made -l take optional filename.  Without filename maintains existing behavior.
4. Since sighup can interupt the select system call, now checking select error code for interupted syscal and logfile enabled.  If those conditions are met, the errorcode is non critical.

Also attempt to address #38 as I was updating signal handling anyways.

### DIFF
--- a/candump.c
+++ b/candump.c
@@ -110,9 +110,19 @@ const int canfd_on = 1;
 const char anichar[MAXANI] = {'|', '/', '-', '\\'};
 const char extra_m_info[4][4] = {"- -", "B -", "- E", "B E"};
 
+#define MAXLOGNAMESZ 100
+FILE *logfile = NULL;
+static char log_filename[MAXLOGNAMESZ];
+
+unsigned char silent = SILENT_INI;
+
 extern int optind, opterr, optopt;
 
 static volatile int running = 1;
+static volatile int flag_reopen_file = 0;
+static volatile unsigned long sighup_count = 0;
+int is_auto_log_name = 0;
+int open_log_file(const char * filename);
 
 void print_usage(char *prg)
 {
@@ -127,7 +137,7 @@ void print_usage(char *prg)
 	fprintf(stderr, "         -a          (enable additional ASCII output)\n");
 	fprintf(stderr, "         -S          (swap byte order in printed CAN data[] - marked with '%c' )\n", SWAP_DELIMITER);
 	fprintf(stderr, "         -s <level>  (silent mode - %d: off (default) %d: animation %d: silent)\n", SILENT_OFF, SILENT_ANI, SILENT_ON);
-	fprintf(stderr, "         -l          (log CAN-frames into file. Sets '-s %d' by default)\n", SILENT_ON);
+	fprintf(stderr, "         -l <name>   (log CAN-frames into file. Sets '-s %d' by default)\n", SILENT_ON);
 	fprintf(stderr, "         -L          (use log file format on stdout)\n");
 	fprintf(stderr, "         -n <count>  (terminate after reception of <count> CAN frames)\n");
 	fprintf(stderr, "         -r <size>   (set socket receive buffer to <size>)\n");
@@ -163,6 +173,15 @@ void print_usage(char *prg)
 void sigterm(int signo)
 {
 	running = 0;
+}
+
+void sighup(int signo)
+{
+	if(signo == SIGHUP)
+	{
+		flag_reopen_file = 1;
+		sighup_count++;
+	}
 }
 
 int idx2dindex(int ifidx, int socket) {
@@ -214,6 +233,100 @@ int idx2dindex(int ifidx, int socket) {
 	return i;
 }
 
+int sprint_auto_filename_format(char * buffer)
+{
+	time_t currtime;
+	struct tm now;
+
+	if (time(&currtime) == (time_t)-1) {
+		perror("time");
+		//running = 0;
+		return 1;
+	}
+
+	localtime_r(&currtime, &now);
+
+	sprintf(buffer, "candump-%04d-%02d-%02d_%02d%02d%02d.log",
+		now.tm_year + 1900,
+		now.tm_mon + 1,
+		now.tm_mday,
+		now.tm_hour,
+		now.tm_min,
+		now.tm_sec);
+
+	fprintf(stderr, "Enabling Logfile '%s'\n", buffer);
+
+	return 0;
+}
+
+//opens file using global filehandle logfile
+int open_log_file(const char * fileName)
+{
+	if (silent != SILENT_ON)
+		fprintf(stderr, "Warning: Console output active while logging!\n");
+
+	const char * fopen_opts = (sighup_count > 0) ? "a" : "w";
+	logfile = fopen(fileName, fopen_opts);
+
+	if (!logfile) {
+		perror("logfile");
+		return 1;
+	}
+
+	return 0;
+}
+
+//flush,close then reopen log
+int reopen_file(FILE * fileHandle)
+{
+	if(!fileHandle)
+		return 1;
+
+	int errcode = 0;
+
+	if(is_auto_log_name == 1)
+	{
+		errcode = sprint_auto_filename_format(&log_filename[0]);
+
+		if(errcode != 0)
+		{
+			running = 0;
+			return 1;
+		}
+	}
+
+	const char * fopen_opts = (sighup_count > 0) ? "a" : "w";
+	logfile = freopen(log_filename,fopen_opts,fileHandle);
+
+	running = (errcode == 0);
+	flag_reopen_file = 0;
+
+	return 0;
+}
+
+void get_logname_arg(const char * arg)
+{
+	if(arg != 0)
+	{
+		size_t len = strnlen(arg,MAXLOGNAMESZ);
+		if(len > 0 && len < MAXLOGNAMESZ)
+		{
+			strncpy(log_filename,arg,MAXLOGNAMESZ-1);
+		}
+		else
+		{
+			//print_usage(basename(argv[0]));
+			exit(1);
+		}
+		is_auto_log_name = 0;
+	}
+	else
+	{
+		is_auto_log_name = 1;	
+		sprint_auto_filename_format(&log_filename[0]);
+	}
+}
+
 int main(int argc, char **argv)
 {
 	fd_set rdfs;
@@ -223,7 +336,7 @@ int main(int argc, char **argv)
 	unsigned char down_causes_exit = 1;
 	unsigned char dropmonitor = 0;
 	unsigned char extra_msg_info = 0;
-	unsigned char silent = SILENT_INI;
+	
 	unsigned char silentani = 0;
 	unsigned char color = 0;
 	unsigned char view = 0;
@@ -247,16 +360,29 @@ int main(int argc, char **argv)
 	struct ifreq ifr;
 	struct timeval tv, last_tv;
 	struct timeval timeout, timeout_config = { 0, 0 }, *timeout_current = NULL;
-	FILE *logfile = NULL;
 
-	signal(SIGTERM, sigterm);
-	signal(SIGHUP, sigterm);
-	signal(SIGINT, sigterm);
+	{
+		sigset_t sig_mask;
+		sigemptyset (&sig_mask);
+
+		struct sigaction sigterm_action = {.sa_mask = sig_mask, .sa_flags = SA_RESTART, .sa_handler = sigterm};
+		sigaction (SIGHUP, &sigterm_action, NULL);
+		sigaction (SIGINT, &sigterm_action, NULL);
+	}
 
 	last_tv.tv_sec  = 0;
 	last_tv.tv_usec = 0;
 
-	while ((opt = getopt(argc, argv, "t:HciaSs:lDdxLn:r:he8T:?")) != -1) {
+	int getoptargc = argc;
+
+	//Since interface is a required argument, we don't need to parse opt for final arg
+	//This enabled the -l option to take an optional filename
+	if(getoptargc > 0)
+	{
+		getoptargc = argc-1;
+	}
+
+	while ((opt = getopt(getoptargc, argv, ":t:HciaSs:l:DdxLn:r:he8T:?")) != -1) {
 		switch (opt) {
 		case 't':
 			timestamp = optarg[0];
@@ -304,8 +430,30 @@ int main(int argc, char **argv)
 			}
 			break;
 
+		case ':': //handle flags with optional values
+			
+			switch (optopt)
+			{
+			case 'l':
+			{
+				log = 1;
+
+				get_logname_arg(optarg);
+				
+				break;
+			}
+			default:
+				fprintf(stderr, "option -%c is missing a required argument\n", optopt);
+				return EXIT_FAILURE;
+			}		
+			
+			break;
+
 		case 'l':
 			log = 1;
+			
+			get_logname_arg(optarg);
+
 			break;
 
 		case 'D':
@@ -361,6 +509,15 @@ int main(int argc, char **argv)
 	if (optind == argc) {
 		print_usage(basename(argv[0]));
 		exit(0);
+	}
+
+	//Configure SIGHUP handler to reopen file if logging
+	{	
+		sigset_t sig_block_mask;
+		sigfillset (&sig_block_mask);
+		struct sigaction usr_action = {.sa_mask = sig_block_mask, .sa_flags = SA_RESTART};
+		usr_action.sa_handler = log ? sighup : sigterm;		
+		sigaction(SIGHUP, &usr_action, NULL);
 	}
 	
 	if (logfrmt && view) {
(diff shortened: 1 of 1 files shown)


candump.c
```diff
@@ -110,9 +110,19 @@ const int canfd_on = 1;
...
+FILE *logfile = NULL;
+static char log_filename[MAXLOGNAMESZ];
+
+unsigned char silent = SILENT_INI;
+
 extern int optind, opterr, optopt;
 
 static volatile int running = 1;
+static volatile int flag_reopen_file = 0;
+static volatile unsigned long sighup_count = 0;
```
marckleinebudde, line 123: no need to initialize statics with `0`

candump.c
```diff
@@ -110,9 +110,19 @@ const int canfd_on = 1;
 const char anichar[MAXANI] = {'|', '/', '-', '\\'};
 const char extra_m_info[4][4] = {"- -", "B -", "- E", "B E"};
 
+#define MAXLOGNAMESZ 100
+FILE *logfile = NULL;
+static char log_filename[MAXLOGNAMESZ];
+
+unsigned char silent = SILENT_INI;
```
marckleinebudde, line 117: `static`

candump.c
```diff
@@ -165,6 +175,15 @@ void sigterm(int signo)
 	running = 0;
 }
 
+void sighup(int signo)
+{
+	if(signo == SIGHUP)
+	{
```
marckleinebudde, line 181: please use kernel coding style
`if (signo == SIGHUP) {`

hartkopp:
Thanks for the reference to PR https://github.com/linux-can/can-utils/pull/189 from @buglight which instantly came into my mind too.
I'm still not really happy with bloating `candump` in such way, as `candump` should be no file-handling-monster with tons of features but a tool that filters, shows and stores CAN frames.
IMO if we implement this kind of multi-chunk file generation we should support either the size based triggers as suggested by @buglight AND the SIGHUP approach from you.
The PR https://github.com/linux-can/can-utils/pull/189 has different commits we can probably squash to a single patch. And once we have a separated function to hand-over the log stream to a new file we might also trigger this hand-over by SIGHUP.
What needs to be discussed is how the filenames are generated. We might add a postfix number to the original timstamp based filename or we generate a new timestamp based filename for each file (simple / fits better to SIGHUP).
I definitely do not like the change of the `-l` option.
Finally we would wait for PR https://github.com/linux-can/can-utils/pull/264 as new base for further developments.

candump.c
```diff
@@ -214,6 +233,100 @@ int idx2dindex(int ifidx, int socket) {
 	return i;
 }
 
+int sprint_auto_filename_format(char * buffer)
+{
```
marckleinebudde, line 237: `(char *buffer)`

candump.c
```diff
@@ -214,6 +233,100 @@ int idx2dindex(int ifidx, int socket) {
...
+		now.tm_hour,
+		now.tm_min,
+		now.tm_sec);
+
+	fprintf(stderr, "Enabling Logfile '%s'\n", buffer);
+
+	return 0;
+}
+
+//opens file using global filehandle logfile
```
marckleinebudde, line 262: no c++ comments please

candump.c
```diff
@@ -214,6 +233,100 @@ int idx2dindex(int ifidx, int socket) {
...
+int open_log_file(const char * fileName)
+{
```
marckleinebudde, line 264: `(const char *file_name)`

candump.c
```diff
@@ -214,6 +233,100 @@ int idx2dindex(int ifidx, int socket) {
...
+	if (silent != SILENT_ON)
+		fprintf(stderr, "Warning: Console output active while logging!\n");
+
+	const char * fopen_opts = (sighup_count > 0) ? "a" : "w";
```
marckleinebudde, line 268: please move the variable declaration to the beginning of the function.

candump.c
```diff
@@ -214,6 +233,100 @@ int idx2dindex(int ifidx, int socket) {
...
+		return 1;
+	}
+
+	return 0;
+}
+
+//flush,close then reopen log
+int reopen_file(FILE * fileHandle)
+{
+	if(!fileHandle)
```
marckleinebudde, line 282: `if (!`

candump.c
```diff
@@ -214,6 +233,100 @@ int idx2dindex(int ifidx, int socket) {
...
+		return 1;
+
+	int errcode = 0;
+
+	if(is_auto_log_name == 1)
```
marckleinebudde, line 287: `if (...) {`

candump.c
```diff
@@ -214,6 +233,100 @@ int idx2dindex(int ifidx, int socket) {
...
+	{
+		errcode = sprint_auto_filename_format(&log_filename[0]);
```
marckleinebudde, line 289: `(log_filename)`

candump.c
```diff
@@ -214,6 +233,100 @@ int idx2dindex(int ifidx, int socket) {
...
+
+		if(errcode != 0)
```
marckleinebudde, line 291: `if (...) {`

candump.c
```diff
@@ -214,6 +233,100 @@ int idx2dindex(int ifidx, int socket) {
...
+		{
+			running = 0;
+			return 1;
+		}
+	}
+
+	const char * fopen_opts = (sighup_count > 0) ? "a" : "w";
+	logfile = freopen(log_filename,fopen_opts,fileHandle);
```
marckleinebudde, line 299: please add a space after each `,`

candump.c
```diff
@@ -214,6 +233,100 @@ int idx2dindex(int ifidx, int socket) {
...
+	if (!logfile) {
+		perror("logfile");
+		return 1;
+	}
+
+	return 0;
+}
+
+//flush,close then reopen log
+int reopen_file(FILE * fileHandle)
```
marckleinebudde, line 280: static

candump.c
```diff
@@ -214,6 +233,100 @@ int idx2dindex(int ifidx, int socket) {
...
+		{
+			running = 0;
+			return 1;
+		}
+	}
+
+	const char * fopen_opts = (sighup_count > 0) ? "a" : "w";
+	logfile = freopen(log_filename,fopen_opts,fileHandle);
+
+	running = (errcode == 0);
```
marckleinebudde, line 301: please return the error if there is one and handle it in the caller

candump.c
```diff
@@ -214,6 +233,100 @@ int idx2dindex(int ifidx, int socket) {
...
+	flag_reopen_file = 0;
+
+	return 0;
+}
+
+void get_logname_arg(const char * arg)
```
marckleinebudde, line 307: `(const char *arg)`

candump.c
```diff
@@ -214,6 +233,100 @@ int idx2dindex(int ifidx, int socket) {
...
+{
+	if(arg != 0)
```
marckleinebudde, line 309: `if (arg) {`

candump.c
```diff
@@ -214,6 +233,100 @@ int idx2dindex(int ifidx, int socket) {
...
+	{
+		size_t len = strnlen(arg,MAXLOGNAMESZ);
+		if(len > 0 && len < MAXLOGNAMESZ)
```
marckleinebudde, line 312: `if (len > 0 && ...) {`

candump.c
```diff
@@ -214,6 +233,100 @@ int idx2dindex(int ifidx, int socket) {
...
+	flag_reopen_file = 0;
+
+	return 0;
+}
+
+void get_logname_arg(const char * arg)
+{
+	if(arg != 0)
+	{
+		size_t len = strnlen(arg,MAXLOGNAMESZ);
```
marckleinebudde, line 311: please add a newline after the variable declaration and add a space after the `,`

candump.c
```diff
@@ -214,6 +233,100 @@ int idx2dindex(int ifidx, int socket) {
...
+		if(len > 0 && len < MAXLOGNAMESZ)
+		{
+			strncpy(log_filename,arg,MAXLOGNAMESZ-1);
+		}
+		else
```
marckleinebudde, line 316: `} else {`

candump.c
```diff
@@ -214,6 +233,100 @@ int idx2dindex(int ifidx, int socket) {
...
+		{
+			//print_usage(basename(argv[0]));
+			exit(1);
```
marckleinebudde, line 319: please return the error and handle it

candump.c
```diff
@@ -214,6 +233,100 @@ int idx2dindex(int ifidx, int socket) {
...
+		}
+		is_auto_log_name = 0;
+	}
+	else
```
marckleinebudde, line 323: `} else {`

candump.c
```diff
@@ -247,16 +360,29 @@ int main(int argc, char **argv)
 	struct ifreq ifr;
 	struct timeval tv, last_tv;
 	struct timeval timeout, timeout_config = { 0, 0 }, *timeout_current = NULL;
-	FILE *logfile = NULL;
 
-	signal(SIGTERM, sigterm);
-	signal(SIGHUP, sigterm);
-	signal(SIGINT, sigterm);
+	{
```
marckleinebudde, line 364: why do you open a new scope here?

marckleinebudde (changes requested):
Please use the kernel coding style:
https://github.com/torvalds/linux/blob/master/Documentation/process/coding-style.rst

candump.c
```diff
@@ -111,18 +111,18 @@ const char anichar[MAXANI] = {'|', '/', '-', '\\'};
...
 
 static volatile int running = 1;
-static volatile int flag_reopen_file = 0;
-static volatile unsigned long sighup_count = 0;
-int is_auto_log_name = 0;
-int open_log_file(const char * filename);
+static volatile int flag_reopen_file;
+static volatile unsigned long sighup_count;
+
+static int is_auto_log_name = 0;
```
marckleinebudde, line 125: no need to initialize static vars with 0

candump.c
```diff
@@ -259,14 +257,13 @@ int sprint_auto_filename_format(char * buffer)
 	return 0;
 }
 
-//opens file using global filehandle logfile
-int open_log_file(const char * fileName)
+/*opens file using global var logfile*/
```
marckleinebudde, line 260: please add a space before and after the `*`

`/* comment */`

candump.c
```diff
@@ -276,55 +273,45 @@ int open_log_file(const char * fileName)
...
-		}
-		else
-		{
-			//print_usage(basename(argv[0]));
-			exit(1);
+	if (arg != 0) {
+		const size_t len = strnlen(arg, MAXLOGNAMESZ);
+
+		if (len > 0 && len < MAXLOGNAMESZ) {
+			strncpy(log_filename, arg, MAXLOGNAMESZ-1);
```
marckleinebudde, line 304: `MAXLOGNAMESZ - 1`

candump.c
```diff
@@ -511,15 +500,13 @@ int main(int argc, char **argv)
...
-	//Configure SIGHUP handler to reopen file if logging
-	{	
-		sigset_t sig_block_mask;
-		sigfillset (&sig_block_mask);
-		struct sigaction usr_action = {.sa_mask = sig_block_mask, .sa_flags = SA_RESTART};
-		usr_action.sa_handler = log ? sighup : sigterm;		
-		sigaction(SIGHUP, &usr_action, NULL);
-	}
-	
+	/*Configure SIGHUP handler to reopen file if logging*/
```
marckleinebudde, line 503: `/* add space */`

candump.c
```diff
@@ -511,15 +500,13 @@ int main(int argc, char **argv)
...
+	sigset_t sig_block_mask;
+	sigfillset(&sig_block_mask);
+	struct sigaction usr_action = { .sa_mask = sig_block_mask, .sa_flags = SA_RESTART };
```
marckleinebudde, line 506: please move the variable declaration to the beginning of the scope

candump.c
```diff
@@ -962,10 +942,8 @@ int main(int argc, char **argv)
 			fflush(stdout);
 		}
 
-		if(flag_reopen_file == 1)
-		{
-			if(reopen_file(logfile) != 0)
-			{
+		if (flag_reopen_file == 1) { 
+			if(reopen_file(logfile) != 0) {
```
marckleinebudde, line 946: `if (`

marckleinebudde (changes requested):
Please fix the remaining coding style and then squash all patches, that there is only one.

candump.c
hartkopp: There are general issues with this patch. See https://github.com/linux-can/can-utils/pull/268#issuecomment-739410309
It does not make sense to discuss style problems until the general approach and concept is under discussion.

marckleinebudde: > IMO if we implement this kind of multi-chunk file generation we should support either the size based triggers as suggested by @BugLight AND the SIGHUP approach from you.

What do you mean: "either...or" (as in one or the other) or "and" (as in both)?

Once file naming and new file generation is solved, triggering the switch to a new file due to `SIGHUP` or size limit is hopefully not that complicated.

hartkopp: Yes, this was my intention!
But the way the filename is generated still needs some discussion IMO.
And finally I thought the PR https://github.com/linux-can/can-utils/pull/189 to be a simpler starting point, from which we might add the SIGHUP handling. But that's probably just my view.

BugLight:
@hartkopp, for me, it seems that name files using only timestamp (no part number) is more general and better approach.

richnetdesign:
Thanks for the feedback everybody.  I have rebased on the recent commits to master, and most importantly the epoll changes.

Regarding file name generation I want the ability to choose the output filename/format.
Of course I also want to support auto formatted names for convenience and backwards compatibility.  The path of least resistance approach I initially took was making `-l` take an optional file name.  I would be open to discussing other options, such as a separate parameter for log file name / format.

Below is a simplified example of a logrotate setup that I am looking to support.  When triggered, logrotate copies that static filename to configured name / path, then sends SIGHUP.  The idea is that previous data is archived, and candump recreates the original filename and continues logging.

/etc/logrotate.d/candump1
```
/tmp/candump1.log {
    hourly
    missingok
    compress
    maxsize 1K
    su root root
    rotate 50
    dateext
    dateformat -%Y%m%d_%H.%M.%s
    postrotate
        /usr/bin/kill -HUP `cat /var/run/candump1.pid 2> /dev/null`
    endscript
}
```

`./candump -l /tmp/candump1.log vcan0 & echo $! > /var/run/candump1.pid`

sudo logrotate /etc/logrotate.conf

ls /tmp/candump1*

hartkopp:
> When triggered, logrotate copies that static filename to configured name / path, then sends SIGHUP. The idea is that previous data is archived, and candump recreates the original filename and continues logging.

This sounds strange! You COPY the file and THEN issue the SIGHUP? Isn't there the chance that you record some frames twice?

This code makes a very clear close/open sequence between two CAN frames:
https://github.com/linux-can/can-utils/pull/189/commits/21f4b81510a7ad4cbca5e1eab3598ad75e03f7ac#diff-c9165277a67241a5db54c2140829d4ebda95c35ce668d8ea27832afb24bf2dc5R779

> `./candump -l /tmp/candump1.log vcan0 & echo $! > /var/run/candump1.pid`

Instead of your suggestion `candump -l filename [other options]` you can also run `candump -L [other options] > filename` :-)

richnetdesign:
Sorry I meant move/rename the file.  This is the debug output of related logrotated action.
```
renaming /tmp/candump1.log to /tmp/candump1.log-20201206_16.51.1607291480
creating new /tmp/candump1.log mode = 0664 uid = 1000 gid = 1000
running postrotate script
running script with args /tmp/candump1.log /tmp/candump1.log-20201206_16.51.1607291480: "
        /usr/bin/kill -HUP `cat /var/run/candump1.pid 2> /dev/null`
"
```

richnetdesign:
> you can also run candump -L [other options] > filename :-)
It is possible to do something like this, but with >> always append mode and logrotated, but it is less ideal, as it is possible to lose data.  This is because it requires the "copytruncate" option, which copies the file, and truncates it so the original filehandle is preserved.  
In contrast the example I provided integrates with candump to move then close the file handle, and makes a new replacement file handle pointing to the original log location.  This ensures a clean hand off from one file to another.

I agree that logrotate is kind of weird.  However, it is widely deployed so building in compatibility would be helpful.  I like the idea of unifying the internal rotation functionality to internally use SIGHUP to trigger the size based rotation events.

> This code makes a very clear close/open sequence between two CAN frames:
> 21f4b81#diff-c9165277a67241a5db54c2140829d4ebda95c35ce668d8ea27832afb24bf2dc5R779

Since a SIGHUP can occur at any time (even when can messages aren't being received), I am currently setting a flag on the signal and handling the rotation in the main loop.  I also needed to handle the error result from interrupting epoll with sighup signal, so i also rotate/reopen file immediately if this occurs.

candump.c
```diff
@@ -314,9 +429,32 @@ int main(int argc, char **argv)
...
+			break;
 
 		case 'l':
 			log = 1;
+			
+			if (process_logname_arg(optarg) != 0) {
+				is_auto_log_name = 0;
+				print_usage(basename(argv[0]));
+				exit(1);
+			}
```
hartkopp, line 457: Can you please think about an integration into a logrotated configuration, that can cope with the assumptions:

1. The option '-l' remains untouched
2. The logfile is **closed** and a **new logfile is opened** with a **new data-generated filename**

So no re-open/append stuff and no change with the naming.
You might also execute candump in `/tmp/candump/` to place the files there for logrotation post-processing.

I'm fine with creating two file splitting triggers about the length and the SIGHUP. But the current approach just adds too much complexity that does not need to be inside candump.

jayceslesar:
+1 on getting this in

hartkopp:
The infrastructure to create individual filenames changed in https://github.com/linux-can/can-utils/commit/ad250a68dc838824f6e986ac352c8d0fc52cea8e and https://github.com/linux-can/can-utils/commit/c70d0a8e61140942447ef75b0afc2662551a4fad which solved my main concern to not break the `-l` command line option.

So the original patch needs to be reworked and rebased.

Additionally it has to be discussed whether the logfile names are autogenerated with date/time or should be based on an individual filename with an increasing counter (see https://github.com/linux-can/can-utils/pull/268#issuecomment-739509365).

PR #285, opened 2021-04-14
Update description for candump

kurnev:
Hello everyone, 

This may sound silly, but I just spend two hours debugging why this `-T` option won't terminate the process after specified timeout. Later I figured out it only terminates if nothing was received. A small change, but probably it saves some time to someone else.


marckleinebudde:
Thanks

PR #286, opened 2021-04-16
asc2log: Correct usec overflow handling

simontegelid:
This commit fixes this sort of lines
(1614188635.1000000) can1 4A8##28001E9A318ACC0 R
to become this
(1614188636.000000) can1 4A8##28001E9A318ACC0 R

Canplayer was barfing about the incorrect lines.

### DIFF
--- a/asc2log.c
+++ b/asc2log.c
@@ -119,7 +119,7 @@ void calc_tv(struct timeval *tv, struct timeval *read_tv,
 		tv->tv_usec += read_tv->tv_usec;
 	}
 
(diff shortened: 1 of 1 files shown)


hartkopp:
Good catch! Many thanks for the fix!!

hartkopp:
There was a second problem with this check:
https://github.com/linux-can/can-utils/commit/553f0d282b9b93ced1fb372aebdaf812441340ae
Thanks Simon!

Issue #289, opened 2021-05-02
Circular buffer candump feature request


memetb:
Hello,

it would be useful in a project I'm working on to have the `candump` utility output to a circular buffer. 

This feature would be implemented as follows:

- only available in conjunction with `-l` option (file output) OR a dedicated option which semantically forces file output
- circular buffer would not pre-allocate buffer, however, once max sample count reached, would `seek` to the top of the file and continue outputting
- to make things work, output will have to be of 25-wide fixed length whitespace right-padding (c.f. below)
- maybe add options for `sync`

```
(1619926775.202098) can1 1EFFBFDF#0200200300000000
(1619926775.202336) can1 0AB#928E
                         ^         25 wide       ^
```

In my project, this feature is useful for crash-dumping complex systems where there might be multiple vendor products interacting with each other that otherwise can't be captured using traditional logging - the main impediment being the sheer volume of traffic in transit.

I am already investigating a fork of the project, but wondering if there is interest in this feature from the maintainers of `linux-can`, and whether there are architectural decisions which you would like to be in charge of.

Thank you for your time.

PR #290, opened 2021-05-02
Circular buffer file output

memetb:
Circular buffer implementation as per #289.

Sometimes the need arises to log accidents like a dash-cam. High bandwidth CANbus logging becomes expensive from storage a standpoint especially on embedded devices.

This feature allows for ring-buffer type logging where a fixed number of "most recent" entries is constantly on file and ready to be stored in the event of a failure for forensic analysis.

This implementation relies heavily on the existing `log` code-path and as such introduces a (perhaps undesirable) coupling. However, it also avoided trying to make the actual `log` output code-path (line 832) be conditional, and instead outputs in a dedicated code block on line 807.

### DIFF
--- a/candump.c
+++ b/candump.c
@@ -135,7 +135,7 @@ static void print_usage(char *prg)
 	fprintf(stderr, "         -S          (swap byte order in printed CAN data[] - marked with '%c' )\n", SWAP_DELIMITER);
 	fprintf(stderr, "         -s <level>  (silent mode - %d: off (default) %d: animation %d: silent)\n", SILENT_OFF, SILENT_ANI, SILENT_ON);
 	fprintf(stderr, "         -l          (log CAN-frames into file. Sets '-s %d' by default)\n", SILENT_ON);
-	fprintf(stderr, "         -b <count>  (log CAN-frames into a ring buffer file of at most <count> CAN frames)\n");
+	fprintf(stderr, "         -b <count>  (similar to '-l', but using a ring buffer file of at most <count> CAN frames)\n");
 	fprintf(stderr, "         -L          (use log file format on stdout)\n");
 	fprintf(stderr, "         -n <count>  (terminate after reception of <count> CAN frames)\n");
 	fprintf(stderr, "         -r <size>   (set socket receive buffer to <size>)\n");
--- a/candump.c
+++ b/candump.c
@@ -316,6 +316,8 @@ int main(int argc, char **argv)
 	int timeout_ms = -1; /* default to no timeout */
 	FILE *logfile = NULL;
   int buffer_length = 0; /*  number of entries to store in ring buffer */
+  int buffer_index = 0;
+  int line_length = 0; /*  cached length of entries  */
   unsigned char circular = 0;
 
 
@@ -455,7 +457,7 @@ int main(int argc, char **argv)
   }
 
 	if (silent == SILENT_INI) {
-		if (log) {
+		if (log || circular) {
 			fprintf(stderr, "Disabled standard output while logging.\n");
 			silent = SILENT_ON; /* disable output on stdout */
 		} else
@@ -631,7 +633,7 @@ int main(int argc, char **argv)
 			}
 		}
 
-		if (timestamp || log || logfrmt) {
+		if (timestamp || log || circular || logfrmt) {
 			if (hwtimestamp) {
 				const int timestamping_flags = (SOF_TIMESTAMPING_SOFTWARE |
 								SOF_TIMESTAMPING_RX_SOFTWARE |
@@ -669,7 +671,7 @@ int main(int argc, char **argv)
 		}
 	}
 
-	if (log) {
+	if (log || circular) {
 		time_t currtime;
 		struct tm now;
 		char fname[83]; /* suggested by -Wformat-overflow= */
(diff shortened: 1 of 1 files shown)


hartkopp:
Hi Memet,

thanks for your contribution! IMHO candump should stay a command line tool to watch and simply dump CAN content - and if you take a look on the command line options they already got bloated.

Creating a circular buffer as some kind of 'crash recorder' is an interesting use case. But I would see a small separate tool there.

E.g. 
1. read raw CAN frame structures into a circular buffer in some malloc'ed RAM (as binary CAN FD structs)
2. Get a trigger (either over CAN or as process signal using a signal handler)
3. dump the circular buffer to a file (using the functions from lib.h)
4. goto 1

I'm pretty sure this is a more efficient approach than constantly creating ASCII CAN frames and fiddling with the filesystem.

memetb:
Hi Oliver,

no  problem for keeping it as a separate project.

Fyi, storing a CAN circular buffer in memory defeats the purpose of having a forensic crash dump. Fiddling with the file system *is* entirely the point of a transactional record of what the last moments of a component right before a potentially catastrophic failure.

olerem:
I would prefer to have it as separate tool, not separate project. It has some advantages to maintain all the tools  under the hood of linux-can project.

hartkopp:
> Fyi, storing a CAN circular buffer in memory defeats the purpose of having a forensic crash dump. Fiddling with the file system _is_ entirely the point of a transactional record of the last moments of a component right before a potentially catastrophic failure.

Hm, good point. But a circular buffer might not only be interesting for crashes but also for user defined triggers that e.g. would like to have the last 30 secs prior to a specific bitflip in CAN data. The trigger intelligence should not be part of this tool - but waiting for a signal and then dumping the CAN frames to the file system with an actual filename seems a valid non-forensic use case too.

memetb:
@olerem I've tried to make the code-path leverage the existing `-l` flag as much as possible while maintaining the remaining structure of the tool. Making this a separate tool under the same project could be useful from a user experience standpoint, but I'd essentially be duplicating the entire codebase to achieve that. I'm not sure that's going to be desirable from the maintenance standpoint.

@hartkopp: I've already implemented a systemd service on my systems that use signaling: `systemctl reload my-circular-buffer`  is tied to a SIGHUP and restart. In other words, I think the signal idea can be achieve through normal unix mechanisms already by simply restarting the process.

Part of why I like the `candump` tool is that it's extremely fast to launch (and relaunch) even on heavily resource constrained embedded systems.

Open to ideas on how to approach both issues.

olerem:
@memetb , cloning code to a separate project make no sense too.  There are already tools with shared code base. For example this is the shared code https://github.com/linux-can/can-utils/blob/master/libj1939.c for some of j1939 related tools

memetb:
> 
> 
> @memetb , cloning code to a separate project make no sense too. There are already tools with shared code base. For example this is the shared code https://github.com/linux-can/can-utils/blob/master/libj1939.c for some of j1939 related tools

@olerem: not sure I'm understanding what you're saying: you want it *not* to be cloned right? I agree with that. I'm just saying if I were to make this a dedicated tool in the can-utils project, I would end up copying most of `candump.c` to make it happen.

As it stands, I've only added a command line option and essentially what amounts to an `fseek`.

olerem:
Yes, currently you added only one change, but the tool for this use case will need more changes. For example:
- payload filter
- packet based trigger
- ECC error correction
- compression
- some kind of helpers to integrate it with pstore
My point is, this tool is useful and will need more option which will make candump over complicated for the candump use case.

hartkopp:
> @hartkopp: I've already implemented a systemd service on my systems that use signaling: `systemctl reload my-circular-buffer` is tied to a SIGHUP and restart. In other words, I think the signal idea can be achieve through normal unix mechanisms already by simply restarting the process.
> 
> Part of why I like the `candump` tool is that it's extremely fast to launch (and relaunch) even on heavily resource constrained embedded systems.

But restarting the process - even if it launches very fast - would lead to some missing frames. And what if the next trigger occurs earlier than the dump-buffer is filled completely and we probably lose exactly the content we were searching for?

memetb:
@olerem I see. That is fine by me, but I very likely won't 

> But restarting the process - even if it launches very fast - would lead to some missing frames.

Good point. `SIGUSR` is probably the way to go for recycling a log file.

With regards to crash-dumping, I do not expect there to be a race condition with the trigger condition: the whole purpose is to have the system die with the data already there.

memetb:
@olerem, @hartkopp: How does everyone want to proceed on this feature? Is there a team that makes design decisions?

I don't feel comfortable putting the work in and submitting a PR for a *whole* other tool.

memetb:
fyi: @olerem @hartkopp : I have been using this tool (from our internal forked project) for almost a year in the field now. It is a valuable addition to the toolchain. I would recommend reconsidering a path to integration.

olerem:
hi @memetb , sounds good! Please send patches.

memetb:
@olerem not sure what you mean: the patch is implemented in this PR. The outstanding issue is what direction to go in to implement the change. As per my previous comment in may:

> I don't feel comfortable putting the work in and submitting a PR for a whole other tool without an **prior guidance and buy-in.**

PR #294, opened 2021-05-08
Add CANFD_FDF flag

kayoub5:
First commit was taken as is from @hartkopp https://lore.kernel.org/linux-can/20170411134343.3089-1-socketcan@hartkopp.net/ so he was marked as author.

Second commit is the result of agreement on having flags for both CAN and CAN-FD as per https://lore.kernel.org/linux-can/20210503100810.cacbmdfmpjipgoka@pengutronix.de/

### DIFF
--- a/cangw.c
+++ b/cangw.c
@@ -264,7 +264,7 @@ void print_usage(char *prg)
 	fprintf(stderr, "  <instruction>  is one of 'AND' 'OR' 'XOR' 'SET'\n");
 	fprintf(stderr, "  <canfd_frame-elements>  is _one_ or _more_ of 'I'd 'F'lags 'L'ength 'D'ata\n");
 	fprintf(stderr, "  <can_id>  is an u32 value containing the CAN FD Identifier\n");
-	fprintf(stderr, "  <flags>  is an u8 value containing CAN FD flags (CANFD_BRS, CANFD_ESI)\n");
+	fprintf(stderr, "  <flags>  is an u8 value containing CAN FD flags (CANFD_BRS, CANFD_ESI, CANFD_FDF)\n");
 	fprintf(stderr, "  <len>  is an u8 value containing the data length in hex (0 .. 40)\n");
 	fprintf(stderr, "  <can_data>  is always 64(!) u8 values containing the CAN FD frames data\n");
 	fprintf(stderr, "The max. four modifications are performed in the order AND -> OR -> XOR -> SET\n");
--- a/include/linux/can.h
+++ b/include/linux/can.h
@@ -135,9 +135,18 @@ struct can_frame {
  * controller only the CANFD_BRS bit is relevant for real CAN controllers when
  * building a CAN FD frame for transmission. Setting the CANFD_ESI bit can make
  * sense for virtual CAN interfaces to test applications with echoed frames.
+ *
+ * The struct can_frame and struct canfd_frame intentionaly share the same
+ * layout to be able to write CAN frame content into a CAN FD frame structure.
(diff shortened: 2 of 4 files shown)


hartkopp:
Changing the contents of can.h goes via the big loop of Linux mainline development.
Therefore the discussion needs to be continued on the CAN mailing list.
Some more comments on the patches in-line ...

include/linux/can.h
```diff
@@ -114,14 +114,14 @@ struct can_frame {
 		__u8 len;
 		__u8 can_dlc; /* deprecated */
 	};
-	__u8 __pad; /* padding */
+	__u8 flags; /* additional flags */
```
hartkopp, line 117: Why? The flags are only relevant for CAN FD frames.

cangw.c
```diff
@@ -264,7 +264,7 @@ void print_usage(char *prg)
 	fprintf(stderr, "  <instruction>  is one of 'AND' 'OR' 'XOR' 'SET'\n");
 	fprintf(stderr, "  <canfd_frame-elements>  is _one_ or _more_ of 'I'd 'F'lags 'L'ength 'D'ata\n");
 	fprintf(stderr, "  <can_id>  is an u32 value containing the CAN FD Identifier\n");
-	fprintf(stderr, "  <flags>  is an u8 value containing CAN FD flags (CANFD_BRS, CANFD_ESI)\n");
+	fprintf(stderr, "  <flags>  is an u8 value containing CAN FD flags (CANFD_BRS, CANFD_ESI, CANFD_FDF)\n");
```
hartkopp, line 267: This only makes sense when the flag has landed in the mainline kernel.
Btw. I'm not sure if this bit should have a meaning inside the Linux kernel anyway.
The intention is a usage to help user space programs to handle CAN/CANFD frames inside struct canfd_frames.

marckleinebudde: We might set the CANFD_FDF flag inside the alloc_canfd_skb() function....

https://github.com/torvalds/linux/blob/master/drivers/net/can/dev/skb.c#L210

hartkopp: Yes. But we then introduce a flag which has not been set since CAN FD has been supported in Linux 3.6.
Don't get me wrong: I'm not against setting this flag. But we might run into a problem, when user-space programs *rely* on this bit. Therefore my use-case was to 'define and reserve' this bit for user space programs that like to use the canfd_frame structure for dual-use (CAN/CANFD). These programs then know, that they can rely on this bit - independent from the kernel version.

hartkopp: Btw. we might use that bit for kernel internal use cases and improvements too. Why not?
But it has to be clear that user space can not rely on this.

hartkopp: Would you consider to go further with that patch from 2017?
https://lore.kernel.org/linux-can/20170411134343.3089-1-socketcan@hartkopp.net/ 
At least it still applies with an offset of 11 lines ;-)

hartkopp:
The commit 025468842212 ("can: uapi: introduce CANFD_FDF flag for mixed content in struct canfd_frame") made it into upsteam:
https://git.kernel.org/pub/scm/linux/kernel/git/netdev/net-next.git/commit/?id=02546884221279da2725e87e35348290470363d7
And the patch makes clean when the CANFD_FDF flag should be used.

Therefore updated our local includes:
https://github.com/linux-can/can-utils/commit/affdc1b79973c7497bb8607603c24734e11a91aa

PR can be closed.

Thanks @kayoub5 @marckleinebudde for the discussion!

PR #311, opened 2021-08-20
J1939 rx state support

olerem:
Latest net-next kernel is already supporting RX notifications. This patches are extending j1939cat to make use of this new UAPI.


j1939cat.c
```diff
@@ -231,14 +256,39 @@ static int j1939cat_extract_serr(struct j1939cat_priv *priv)
...
+		switch (serr->ee_info) {
+		case J1939_EE_INFO_TX_ABORT:
+			j1939cat_print_timestamp(priv, "TX ABT", &tss->ts[0]);
+			warnx("serr: tx error: %i, %s", serr->ee_errno,
+			      strerror(serr->ee_errno));
+			return serr->ee_errno;
+		case J1939_EE_INFO_RX_RTS:
+			stats->tskey = serr->ee_data;
+			j1939cat_print_timestamp(priv, "RX RTS", &tss->ts[0]);
+			fprintf(stderr, "  total size: %u, pgn=0x%05x, sa=0x%02x, da=0x%02x src_name=0x%08llx, dst_name=0x%08llx)\n",
```
hartkopp, line 268: Hi @olerem ,

my compiler throws two warnings about the `%08llx` specifier here:
`warning: format ‘%llx’ expects argument of type ‘long long unsigned int’, but argument 8 has type ‘uint64_t’ {aka ‘long unsigned int’}`

So it should be  `%08lx`  instead of  `%08llx` , right?

PR #321, opened 2021-11-14
Fixed some undefined behavior and added CodeQL workflow

weichslgartner:
### Issue
```
mkdir build && cd build
CC=clang cmake -DCMAKE_C_FLAGS="-fsanitize=address,undefined" -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=address,undefined" .. && cmake --build .
```
```
$echo "(0.0) can1 ffffffff#00000000" | ./log2long 
can-utils/lib.c:187:23: runtime error: left shift of 15 by 28 places cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /can-utils/lib.c:187:23 in 
(0.0)  can1  3FFFFFFF   [4]  remote request
```

### Proposed fix

cast to unsigned value
`cf->can_id |= ((canid_t)tmp << (7-i)*4);`

### Issue
```
mkdir build && cd build
CC=clang cmake -DCMAKE_C_FLAGS="-fsanitize=address,undefined" -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=address,undefined" .. && cmake --build .
```
```
$echo "echo "0.0000 0  Rx   d 8 8D 00 10 01 00 82 01 00                                                                                    
      0.200000000000000000 0- 0000  Rx   d 8 8D 00 10 01 00 82 01 00" | ./asc2log
can-utils/asc2log.c:100:20: runtime error: signed integer overflow: 200000000000000000 * 100 cannot be represented in type 'long'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior asc2log.c:100:20 in 
```

### Proposed fix
Add overflow check before performing the multiplication
```
if (dplace == 4 && read_tv.tv_usec >= INT_MAX/100)
    return;
```

### Issue:
CodeQL found four  occurence of "potentially overflowing call to snprintf" (_The return value of a call to snprintf is the number of characters that would have been written to the buffer assuming there was sufficient space. In the event that the operation reaches the end of the buffer and more than one character is discarded, the return value will be greater than the buffer size._)

### Proposed fix
Check return value of snprintf:
```
tmp_n = snprintf(buf + n, len - n, ",");
if (tmp_n < 0 || tmp_n >= len - n){
	return n;
}
n += tmp_n;
```

### DIFF
--- a/asc2log.c
+++ b/asc2log.c
@@ -50,6 +50,7 @@
 #include <sys/time.h>
 #include <time.h>
 #include <unistd.h>
+#include <limits.h>
 
 #include <linux/can.h>
 #include <linux/can/error.h>
@@ -206,6 +207,13 @@ void eval_can(char* buf, struct timeval *date_tvp, char timestamps, char base, i
 		if (strlen(dir) != 2) /* "Rx" or "Tx" */
 			return;
 
+		/* check for signed integer overflow */
+		if (dplace == 4 && read_tv.tv_usec >= INT_MAX / 100)
+			return;
+
+		if (dplace == 5 && read_tv.tv_usec >= INT_MAX / 10)
+			return;
+
 		if (dir[0] == 'R')
 			extra_info = " R\n";
 		else
@@ -267,6 +275,14 @@ void eval_canfd(char* buf, struct timeval *date_tvp, char timestamps, int dplace
 		return;
 
 	if (strlen(dir) != 2) /* "Rx" or "Tx" */
+		return;
+
+	/* check for signed integer overflow */
+	if (dplace == 4 && read_tv.tv_usec >= INT_MAX / 100)
+		return;
+
+	/* check for signed integer overflow */
+	if (dplace == 5 && read_tv.tv_usec >= INT_MAX / 10)
 		return;
 
 	if (dir[0] == 'R')
--- a/lib.c
+++ b/lib.c
@@ -568,10 +568,20 @@ static int snprintf_error_data(char *buf, size_t len, uint8_t err,
 
 	for (i = 0; i < arr_len; i++) {
 		if (err & (1 << i)) {
-			if (count)
-				n += snprintf(buf + n, len - n, ",");
-			n += snprintf(buf + n, len - n, "%s", arr[i]);
-			count++;
+			int tmp_n = 0;
+			if (count){
+				/* Fix for potential buffer overflow https://lgtm.com/rules/1505913226124/ */
+				tmp_n = snprintf(buf + n, len - n, ",");
+				if (tmp_n < 0 || tmp_n >= len - n){
+					return n;
+				}
+				n += tmp_n;
+			}
+			tmp_n = snprintf(buf + n, len - n, "%s", arr[i]);
+			if (tmp_n < 0 || tmp_n >= len - n){
+				return n;
+			}
+			n += tmp_n;
 		}
 	}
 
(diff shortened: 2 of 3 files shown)


asc2log.c
```diff
@@ -264,6 +272,10 @@ void eval_canfd(char* buf, struct timeval *date_tvp, char timestamps, int dplace
 	if (strlen(dir) != 2) /* "Rx" or "Tx" */
 		return;
 
+	/* check for signed integer overflow */
+	if (dplace == 4 && read_tv.tv_usec >= INT_MAX/100)
+		return;
+
```
hartkopp, line 278: No check for dplace == 5 as in eval_can() ?

asc2log.c
```diff
@@ -412,7 +424,7 @@ int main(int argc, char **argv)
 	static char base; /* 'd'ec or 'h'ex */
 	static char timestamps; /* 'a'bsolute or 'r'elative */
 	int opt;
-
+	//while (__AFL_LOOP(10000)){
```
hartkopp: Debug leftover?

lib.c
```diff
@@ -184,6 +184,8 @@ int parse_canframe(char *cs, struct canfd_frame *cf) {
 		for (i=0; i<8; i++){
 			if ((tmp = asc2nibble(cs[i])) > 0x0F)
 				return 0;
+
+			/* casting tmp to canid_t to perform left shift of 28 */
```
hartkopp, line 188: Would it make sense to make tmp a canid_t ?

asc2log.c
weichslgartner: Yes seems that I forgot this.

asc2log.c
weichslgartner: Yes, can be deleted.

asc2log.c
weichslgartner: Should be there now

lib.c
weichslgartner: Done

weichslgartner:
Updated the PR.

weichslgartner:
@hartkopp what is the status here? Can I do anything?

hartkopp:
Oh, sorry about that!
I'm fine with the code changes but I would like to have some feedback from @marckleinebudde who is more experienced in the GitHub CI and probably CodeQL workflows.

marckleinebudde:
You do you add the various workflow file and delete them again?
Can you squash the patches so that each problem is fixed in one patch? E.g. the `can_id` overflow ine one patch the `snprintf` in another, etc...

weichslgartner:
@marckleinebudde  the other workflow files produced too much false positives or needed additional work. That is why I removed them. I adapted now the changes into commits to the specific fixes/additions.

asc2log.c
```diff
@@ -182,6 +183,13 @@ void eval_can(char* buf, struct timeval *date_tvp, char timestamps, char base, i
 		if (strlen(dir) != 2) /* "Rx" or "Tx" */
 			return;
 
+		/* check for signed integer overflow */
+		if (dplace == 4 && read_tv.tv_usec >= INT_MAX/100)
```
marckleinebudde, line 187: please add spaces areound the `/`

marckleinebudde: Does it make sense to move the overflow check into `calc_tv()`?

asc2log.c
```diff
@@ -412,7 +428,6 @@ int main(int argc, char **argv)
 	static char base; /* 'd'ec or 'h'ex */
 	static char timestamps; /* 'a'bsolute or 'r'elative */
 	int opt;
-
```
marckleinebudde, line 415: unrelated change

marckleinebudde (changes requested):

marckleinebudde (changes requested):

marckleinebudde:
> @marckleinebudde the other workflow files produced too much false positives or needed additional work. That is why I removed them.

I see, makes sense.

> I adapted now the changes into commits to the specific fixes/additions.

Now reviewing is easier. Looks almost ready. Can you rebase to the current `master` branch?.

asc2log.c
weichslgartner: added spaces. If check is moved to calc_tv then the struct musst be set to a special value and checked later if time struct is valid, or? Seems to be easier to return before calulating invalid tv.

asc2log.c
weichslgartner: deleted empty line

weichslgartner:
done the changes and rebased to upstream master

lib.c
```diff
@@ -159,7 +159,6 @@ int parse_canframe(char *cs, struct canfd_frame *cf) {
 	int i, idx, dlen, len;
 	int maxdlen = CAN_MAX_DLEN;
 	int ret = CAN_MTU;
-	/* unsigned type to prevent signed integer shifting and undefined behavior */
```
marckleinebudde: This should go into the previous patch.

marckleinebudde (changes requested):

lib.c
marckleinebudde: please squash the new patch

PR #329, opened 2022-01-02
testj1939: fix 64-bit types for some platforms

neheb:
Reverted previous commit and added define to get the same types on all
platforms.


marckleinebudde:
Good work! I've enhanced the commit message a bit and merged your patch: https://github.com/linux-can/can-utils/pull/330

PR #336, opened 2022-01-27
Fix error detection of poll function calls

herbrechtsmeier:
The error detection of the poll function call in cangen and cansequence doesn't exit on timeout and j1939cat wrongly checks the return value and instead of the errno.


cansequence.c
```diff
@@ -216,7 +216,7 @@ static void do_send()
 				}
 
 				err = poll(fds, 1, 1000);
-				if (err == -1 && errno != -EINTR) {
+				if (err == 0 || (err == -1 && errno != -EINTR)) {
```
marckleinebudde, line 219: Can you add a fix `-EINTR` here, too.

marckleinebudde (changes requested):

cangen.c
```diff
@@ -504,7 +504,7 @@ int main(int argc, char **argv)
 			}
 			if (polltimeout) {
 				/* wait for the write socket (with timeout) */
-				if (poll(&fds, 1, polltimeout) < 0) {
+				if (poll(&fds, 1, polltimeout) <= 0) {
```
marckleinebudde, line 507: Can you fix the `-EINTR` here, so that it's not treated as an error.

marckleinebudde (changes requested):

cangen.c
herbrechtsmeier: I have add an extra patch.

PR #342, opened 2022-02-09
isotpsend: add new options to generate PDU traffic load

hartkopp:
To generate high load PDU traffic for kernel testing.

Signed-off-by: Oliver Hartkopp <socketcan@hartkopp.net>

### DIFF
--- a/isotpsend.c
+++ b/isotpsend.c
@@ -71,6 +71,8 @@ void print_usage(char *prg)
 	fprintf(stderr, "         -t <time ns>  (frame transmit time (N_As) in nanosecs)\n");
 	fprintf(stderr, "         -f <time ns>  (ignore FC and force local tx stmin value in nanosecs)\n");
 	fprintf(stderr, "         -D <len>      (send a fixed PDU with len bytes - no STDIN data)\n");
+	fprintf(stderr, "         -l <num>      (send num PDUs - use 'i' for infinite loop)\n");
+	fprintf(stderr, "         -g <usecs>    (wait given usecs before sending a PDU)\n");
 	fprintf(stderr, "         -b            (block until the PDU transmission is completed)\n");
 	fprintf(stderr, "         -S            (SF broadcast mode for functional addressing)\n");
 	fprintf(stderr, "         -L <mtu>:<tx_dl>:<tx_flags>  (link layer options for CAN FD)\n");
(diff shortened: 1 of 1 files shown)


isotpsend.c
```diff
@@ -178,6 +182,22 @@ int main(int argc, char **argv)
 		    }
 		    break;
 
+	    case 'l':
+		    if (optarg[0] == 'i') {
+			    loops = 0; /* infinite loop */
+		    } else {
+			    loops = strtoul(optarg, (char **)NULL, 10);
```
marckleinebudde, line 189: `(char **)` cast not needed

isotpsend.c
```diff
@@ -178,6 +182,22 @@ int main(int argc, char **argv)
...
+			    if (!loops) {
+				    fprintf(stderr, "Invalid argument for option -l !\n");
```
marckleinebudde, line 191: nitpick: No space in front of the `!`?

isotpsend.c
```diff
@@ -178,6 +182,22 @@ int main(int argc, char **argv)
...
+				    return 1;
+			    }
+		    }
+		    break;
+
+	    case 'g':
+		    usecs = strtoul(optarg, (char **)NULL, 10);
```
marckleinebudde, line 198: `(char **)` cast not needed

isotpsend.c
```diff
@@ -87,6 +89,8 @@ int main(int argc, char **argv)
     static struct can_isotp_ll_options llopts;
     int opt;
     extern int optind, opterr, optopt;
+    unsigned int loops = 1; /* one (== no) loop by default */
+    unsigned int usecs = 0; /* wait before sending the PDU */
```
marckleinebudde, line 93: My man page says `usleep()` uses a `useconds_t` as argument.

marckleinebudde (changes requested):

isotpsend.c
hartkopp: Ugh. I have tons(!) of these casts in the can-utils code. Should I create an additional patch to remove them entirely?

isotpsend.c
hartkopp: ack

isotpsend.c
hartkopp: Yes and `useconds_t` is `unsigned long`. I will put `useconds_t` here and let `strtoul()` complain if it doesn't fit ;-)

isotpsend.c
hartkopp: Yes. This is our common coding style in can-utils.

hartkopp: Ah! The parser stopped at `if (!loops)` :-D

Will fix that.

PR #345, opened 2022-02-15
Add optional filename argument for -l

xR3b0rn:


candump.c
```diff
@@ -314,15 +314,16 @@ int main(int argc, char **argv)
...
 
 	signal(SIGTERM, sigterm);
 	signal(SIGHUP, sigterm);
 	signal(SIGINT, sigterm);
 
 	last_tv.tv_sec = 0;
 	last_tv.tv_usec = 0;
-
-	while ((opt = getopt(argc, argv, "t:HciaSs:lDdxLn:r:he8T:?")) != -1) {
+
```
marckleinebudde, line 325: unrelated change

candump.c
```diff
@@ -662,13 +664,23 @@ int main(int argc, char **argv)
...
-			now.tm_min,
-			now.tm_sec);
+		if (!log_filename) {
+			sprintf(fname, "candump-%04d-%02d-%02d_%02d%02d%02d.log",
+				now.tm_year + 1900,
+				now.tm_mon + 1,
+				now.tm_mday,
+				now.tm_hour,
+				now.tm_min,
+				now.tm_sec);
```
marckleinebudde, line 674: assign `log_filename = fname;` here....

candump.c
```diff
@@ -662,13 +664,23 @@ int main(int argc, char **argv)
...
+		} else {
+			int len = strlen(fname);
```
marckleinebudde: ...remove that `else`...and use `log_filename` in `fopen()`instead

candump.c
```diff
@@ -134,7 +134,7 @@ static void print_usage(char *prg)
 	fprintf(stderr, "         -a          (enable additional ASCII output)\n");
 	fprintf(stderr, "         -S          (swap byte order in printed CAN data[] - marked with '%c' )\n", SWAP_DELIMITER);
 	fprintf(stderr, "         -s <level>  (silent mode - %d: off (default) %d: animation %d: silent)\n", SILENT_OFF, SILENT_ANI, SILENT_ON);
-	fprintf(stderr, "         -l          (log CAN-frames into file. Sets '-s %d' by default)\n", SILENT_ON);
+	fprintf(stderr, "         -l <fname>  (log CAN-frames into file. Sets '-s %d' by default)\n", SILENT_ON);
```
hartkopp, line 137: What happens to `candump -l vcan0 vcan1` ??
I don't see, that this way to add an 'optional' filename does not work as it breaks the current behaviour.

xR3b0rn: Yeah I see, `candump -l vcan0 vcan1` would change to `candump -l -- vcan0 vcan1`. So breaking compatibility is no option? Another solution would be to introduce a new depending command line parameter, what would be kind of ugly, I guess... Any other ideas?

hartkopp: Adding options is ok, changing options which might break peoples scripts not.
Even if it looks ugly I would suggest a different option, e.g. `-F <fname> (log into <fname>)` or `-f <fname>` ...

hartkopp: > Another solution would be to introduce a new depending command line parameter, what would be kind of ugly, I guess... Any other ideas perhaps?

What about this suggestion? Would it fit for you?
https://github.com/hartkopp/can-utils/commit/74d9feb1044a0ccae63e93ac828b2fdecc012ae6

xR3b0rn: Looks great.

hartkopp: Thanks @xR3b0rn !
The new option is now applied in master branch.

PR #354, opened 2022-04-12
Return -1 Exit-Code if compare_frame fails

philschenker:
To easily get the errors on an overlying script do return -1 if compare_frame fails


marckleinebudde:
Does `canfdtest` exit in case `compare_frame()` fails?

canfdtest.c
```diff
@@ -117,31 +117,35 @@ static void print_compare(
 	print_frame(rec_id, rec_data, rec_dlc, 0);
 }
 
-static void compare_frame(const struct can_frame *exp, const struct can_frame *rec, int inc)
+static int compare_frame(const struct can_frame *exp, const struct can_frame *rec, int inc)
 {
-	int i;
+	int i, err=0;
```
marckleinebudde, line 122: please add spaces around the `=`

canfdtest.c
```diff
@@ -346,7 +350,7 @@ static int can_echo_gen(void)
 
 			/* own frame */
 			if (rx_frame.can_id == can_id_ping) {
-				compare_frame(&tx_frames[recv_tx_pos], &rx_frame, 0);
+				err = compare_frame(&tx_frames[recv_tx_pos], &rx_frame, 0);
```
marckleinebudde, line 353: Where is the `err` evaluated?

marckleinebudde (changes requested):

canfdtest.c
marckleinebudde: Otherwise looks good.

canfdtest.c
philschenker: yes, and err is just returned, passed on to main() and there returned on the very last line.

philschenker:
I found another path where the return value was ignored, I added a commit for this case.

canfdtest.c
philschenker: done, thanks for your feedback :) !

Issue #369, opened 2022-06-30
comparison of integers of different signs: 'int' and 'unsigned long'


marckleinebudde:
Sorry I didn't have time to dig more into the issue but https://github.com/linux-can/can-utils/pull/321 breaks the Android build:

```
external/can-utils/lib.c:575:28: error: comparison of integers of different signs: 'int' and 'unsigned long' [-Werror,-Wsign-compare]
                                if (tmp_n < 0 || tmp_n >= len - n){
                                                 ~~~~~ ^  ~~~~~~~
external/can-utils/lib.c:581:27: error: comparison of integers of different signs: 'int' and 'unsigned long' [-Werror,-Wsign-compare]
                        if (tmp_n < 0 || tmp_n >= len - n){
                                         ~~~~~ ^  ~~~~~~~
external/can-utils/lib.c:661:28: error: comparison of integers of different signs: 'int' and 'unsigned long' [-Werror,-Wsign-compare]
                                if (tmp_n < 0 || tmp_n >= len - n){
                                                 ~~~~~ ^  ~~~~~~~
external/can-utils/lib.c:667:27: error: comparison of integers of different signs: 'int' and 'unsigned long' [-Werror,-Wsign-compare]
                        if (tmp_n < 0 || tmp_n >= len - n){
                                         ~~~~~ ^  ~~~~~~~
4 errors generated.
```

_Originally posted by @gibsson in https://github.com/linux-can/can-utils/issues/368#issuecomment-1170868223_

weichslgartner:
will take care of it

sisco0:
@weichslgartner I agree with your solution if we should not consider the case that `n` be larger than `len`.
I could not see the comments from the `gh issue view` command, that is the reason why I opened other PR and closed it.

Cheers!

PR #371, opened 2022-07-03
fix: fix comparison types by casting

sisco0:
A cast was introduced to fix comparison types. Now, both sides of the inequality contain signed types and issue #369 should be solved.


sisco0:
Closed, as other PR was proposed before.

PR #393, opened 2023-01-02
candump: always use IFNAMSIZ instead of figuring out the length of the interface names on the fly

marckleinebudde:
...this doesn't work properly if listening on "any" interfaces. The output will be frazzled if a device with a longer name is printed for the first time.

For the log format don't do any indention at all. This keeps log files smaller.

### DIFF
--- a/candump.c
+++ b/candump.c
@@ -108,7 +108,6 @@ static struct if_info sock_info[MAXSOCK];
 static char *progname;
 static char devname[MAXIFNAMES][IFNAMSIZ+1];
 static int dindex[MAXIFNAMES];
-static int max_devname_len; /* to prevent frazzled device name output */
 static const int canfd_on = 1;
 
 #define MAXANI 4
@@ -208,9 +207,6 @@ static int idx2dindex(int ifidx, int socket)
 	ifr.ifr_ifindex = ifidx;
 	if (ioctl(socket, SIOCGIFNAME, &ifr) < 0)
 		perror("SIOCGIFNAME");
-
-	if (max_devname_len < (int)strlen(ifr.ifr_name))
-		max_devname_len = strlen(ifr.ifr_name);
 
 	strcpy(devname[i], ifr.ifr_name);
 
@@ -500,9 +496,6 @@ int main(int argc, char **argv)
 			fprintf(stderr, "name of CAN device '%s' is too long!\n", ptr);
 			return 1;
 		}
-
-		if (nbytes > max_devname_len)
-			max_devname_len = nbytes; /* for nice printing */
 
 		addr.can_family = AF_CAN;
 
@@ -806,9 +799,8 @@ int main(int argc, char **argv)
 
 				/* log CAN frame with absolute timestamp & device */
 				sprint_canframe(buf, &frame, 0, maxdlen);
-				fprintf(logfile, "%s%*s %s%s\n", ts_buf,
-					max_devname_len, devname[idx], buf,
-					extra_info);
+				fprintf(logfile, "%s%s %s%s\n", ts_buf,
+					devname[idx], buf, extra_info);
 			}
 
 			if ((logfrmt) && (silent == SILENT_OFF)){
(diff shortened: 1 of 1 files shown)


marckleinebudde:
Happy new year @hartkopp!

What do you think of the logfile change?

hartkopp:
Hi @marckleinebudde !    Happy new year too!

I wonder if this is really an improvement. The `max_devname_len` mechanic is optimizing the needed space in the lines.
And having all the CAN-IDs and data starting in one column increases the readability too - even when looking into logfiles.

When logfiles are stored they are usually zipped, where it makes no remarkable difference whether there are one or two more spaces after the device name.

So from the usability and readability standpoint I only see regressions in the suggested patch ¯\\_(ツ)_/¯

marckleinebudde:
What about keeping the `max_devname_len` for the log output and using `IFNAMSIZ` for non log output?

marckleinebudde:
This really makes me cry:

```
$ candump any,0:0,#FFFFFFFF -cexdtA -H
 (1970-01-01 01:00:00.000000)  flexcan0  RX - -  002   [8]  00 00 00 00 00 00 00 00
 (2023-01-03 12:02:51.803214)  mcp251xfd0  TX - -  002   [8]  00 00 00 00 00 00 00 00
 (1970-01-01 01:00:00.000000)    flexcan0  RX - -  002   [8]  01 00 00 00 00 00 00 00
 (2023-01-03 12:02:52.003533)  mcp251xfd0  TX - -  002   [8]  01 00 00 00 00 00 00 00
 (1970-01-01 01:00:00.000000)    flexcan0  RX - -  002   [8]  02 00 00 00 00 00 00 00
 (2023-01-03 12:02:52.203833)  mcp251xfd0  TX - -  002   [8]  02 00 00 00 00 00 00 00
```

hartkopp:
And this would make me cry:

```
$ candump any,0:0,#FFFFFFFF -cexdtA -H
 (1970-01-01 01:00:00.000000)              can0  RX - -  002   [8]  00 00 00 00 00 00 00 00
 (2023-01-03 12:02:51.803214)              can1  TX - -  002   [8]  00 00 00 00 00 00 00 00
 (1970-01-01 01:00:00.000000)              can0  RX - -  002   [8]  01 00 00 00 00 00 00 00
 (2023-01-03 12:02:52.003533)              can1  TX - -  002   [8]  01 00 00 00 00 00 00 00
 (1970-01-01 01:00:00.000000)              can0  RX - -  002   [8]  02 00 00 00 00 00 00 00
 (2023-01-03 12:02:52.203833)              can1  TX - -  002   [8]  02 00 00 00 00 00 00 00
```

I would prefer this instead:
```
$ candump any,0:0,#FFFFFFFF -cexdtA -H
 (1970-01-01 01:00:00.000000)  can0  RX - -  002   [8]  00 00 00 00 00 00 00 00
 (2023-01-03 12:02:51.803214)  can1  TX - -  002   [8]  00 00 00 00 00 00 00 00
 (1970-01-01 01:00:00.000000)  can0  RX - -  002   [8]  01 00 00 00 00 00 00 00
 (2023-01-03 12:02:52.003533)  can1  TX - -  002   [8]  01 00 00 00 00 00 00 00
 (1970-01-01 01:00:00.000000)  can0  RX - -  002   [8]  02 00 00 00 00 00 00 00
 (2023-01-03 12:02:52.203833)  can1  TX - -  002   [8]  02 00 00 00 00 00 00 00
```

marckleinebudde:
I have predictable interface names, so no `can0` and `can1` :)

hartkopp:
> I have predictable interface names, so no `can0` and `can1` :)

Yes. But we have to take care for all use-cases and naming schemes. E.g. when I have a setup in a vehicle I usually rename the CAN interfaces according to their real function, e.g. `pwr` for Powertrain or `kom` for Komfort-CAN which also helps to identify and map the CAN data bases. Otherwise it is usually hard to retrieve the CAN bus assignment afterwards,

As you already moved away from the standard/common `canX` naming scheme in your setup I would suggest to rename the interfaces in a way that they share the same name length, if you really bother at increasing the name length one or two times.

Always printing 16 bytes (IFNAMSZ) and having issues to get the interface and CAN data in line with the 'far distant' timestamp is not an improvement IMO. That's not worth it.

PR #396, opened 2023-01-16
Supplying the ifname correctly and removing the now obsolete check of…

t-kopp:
… the length of the ifname. Code taken from cansend.c

This fixes an issue introduced in https://github.com/linux-can/can-utils/commit/1f96d674c0cb7b60ee315981f12a7feb94829697 where fd doesn't work with cangen anymore.

### DIFF
--- a/cangen.c
+++ b/cangen.c
@@ -596,23 +596,23 @@ int main(int argc, char **argv)
 		return 1;
 	}
 
-	if (strlen(argv[optind]) >= IFNAMSIZ) {
-		printf("Name of CAN device '%s' is too long!\n\n", argv[optind]);
-		return 1;
-	}
-
 	s = socket(PF_CAN, SOCK_RAW, CAN_RAW);
 	if (s < 0) {
 		perror("socket");
 		return 1;
 	}
 
-	addr.can_family = AF_CAN;
-	addr.can_ifindex = if_nametoindex(argv[optind]);
-	if (!addr.can_ifindex) {
+	strncpy(ifr.ifr_name, argv[optind], IFNAMSIZ - 1);
+	ifr.ifr_name[IFNAMSIZ - 1] = '\0';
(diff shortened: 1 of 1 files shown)


hartkopp:
Hi,
sorry that I've seen this patch too late but I think this patch does not add any improvement.

In all places in candump we inform the user about CAN interfaces names that are too long - and now we are silently cutting the string without any feedback.

Additionally a memset is introduced for a structure that has been properly initialized before.

I tend to either revert this patch or simply force push the previous version.

t-kopp:
Hi,

The main purpose of this patch was to make FD work again which didn't work
at all anymore because the ifname was missing.

Should the patch be reverted and instead this one applied?

https://github.com/t-kopp/can-utils/commit/90d854049f1413fc9c52e4aed880119a76f70640

I tried to make the handling more uniform in cangen and cansend but
apparently in the wrong direction :)

hartkopp:
Hi Thomas,

not really sure what you mean with 'FD' ?

Maybe this is the missing bit to understand your patch then ;-)

marckleinebudde:
FD as in CAN-FD. I'll force-push-remove Thomas's patch and revert mine https://github.com/linux-can/can-utils/commit/1f96d674c0cb7b60ee315981f12a7feb94829697 instead.

hartkopp:
Sorry but I still feel somehow disconnected. Why has the interface name an impact on whether CAN FD works or not?

t-kopp:
The problem is here:
https://github.com/linux-can/can-utils/blob/master/cangen.c#L636

The now removed patch from Marc removed the ifname= and this line crashed.

hartkopp:
Aah. Got it now! The check for CAN FD uses the ifr struct which was not filled accordingly after this patch https://github.com/linux-can/can-utils/commit/1f96d674c0cb7b60ee315981f12a7feb94829697 

Ok.

PR #401, opened 2023-01-27
cangen: new generation mode - partially randomized payload (-D)

majkk1:
EDIT: `*` symbol was changed to `x`, original message accordingly edited.

During the CAN bus testing, partially fuzzing of frame payload can be useful. This pull request add this feature.

Part of the fixed data payload can be randomized using asterisk (*) symbol. Nibbles marked with asterisk will be randomized.

For example `$ vcan0 -D 11xx3344DEADBEEF -L 8`:
 -> second byte of the payload will be randomized.

### DIFF
--- a/cangen.c
+++ b/cangen.c
@@ -81,6 +81,12 @@
 #define MODE_FIX 2
 #define MODE_RANDOM_EVEN 3
 #define MODE_RANDOM_ODD 4
+#define MODE_RANDOM_FIX 5
+
+#define NIBBLE_H 1
+#define NIBBLE_L 2
+
+#define CHAR_RANDOM 'x'
 
 extern int optind, opterr, optopt;
 
@@ -187,7 +193,8 @@ static void print_usage(char *prg)
 	fprintf(stderr, " 'e'     => random values, even ID\n");
 	fprintf(stderr, " 'o'     => random values, odd ID\n");
 	fprintf(stderr, " 'i'     => increment values\n");
-	fprintf(stderr, " <value> => fixed value (in hexadecimal for -I and -D)\n\n");
+	fprintf(stderr, " <value> => fixed value (in hexadecimal for -I and -D)\n");
+	fprintf(stderr, "         => nibbles written as x are randomized (only -D)\n\n");
 	fprintf(stderr, "The gap value (in milliseconds) may have decimal places, e.g. '-g 4.73'\n");
 	fprintf(stderr, "When incrementing the CAN data the data length code minimum is set to 1.\n");
 	fprintf(stderr, "CAN IDs and data content are given and expected in hexadecimal values.\n\n");
@@ -198,6 +205,8 @@ static void print_usage(char *prg)
 	fprintf(stderr, "\t(generate EFF frames, incr. length)\n");
 	fprintf(stderr, "%s vcan0 -D 11223344DEADBEEF -L 8\n", prg);
 	fprintf(stderr, "\t(fixed CAN data payload and length)\n");
+	fprintf(stderr, "%s vcan0 -D 11xx3344DEADBEEF -L 8\n", prg);
+	fprintf(stderr, "\t(fixed CAN data payload where 2. byte is randomized, fixed length)\n");
 	fprintf(stderr, "%s vcan0 -I 555 -D CCCCCCCCCCCCCCCC -L 8 -g 3.75\n", prg);
 	fprintf(stderr, "\t(generate a fix busload without bit-stuffing effects)\n");
 	fprintf(stderr, "%s vcan0 -g 0 -i -x\n", prg);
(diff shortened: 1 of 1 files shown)


cangen.c
```diff
@@ -81,6 +81,10 @@
 #define MODE_FIX 2
 #define MODE_RANDOM_EVEN 3
 #define MODE_RANDOM_ODD 4
+#define MODE_FORMAT 5
```
hartkopp, line 84: Cool improvement!

Two nitpicks:

`MODE_FORMAT` is very unspecific. What about `MODE_RANDOM_FIX` instead?

cangen.c
```diff
@@ -398,6 +405,32 @@ enum {
 	OPT_START = UCHAR_MAX + 2,
 };
 
+/*
+ * Search for '*' in dataoptarg, save its position, replace it with 0.
```
hartkopp, line 409: I'm not sure if it's a good idea to have an asterisk `*`  here. Can this interfere with the expansion of `*` in shell scripts?

Would `x` also be ok for you?

hartkopp (commented):
Nice feature!

cangen.c
majkk1: Sounds good! I wasn't sure about naming, happy to change it. I used format because I think of it as a format string - maybe in the future, they may be more 'format string' supported but I don't have idea about anything additional useful now.

cangen.c
majkk1: Not sure - any shell expert here? However `*` is usable for example in nmap so I didn't expect any issues and it sounds the most user-friendly for me.

If you want, we can change to `x`.

Additionaly, the formatting character shouldn't be hardcoded, maybe some #define is better?

hartkopp: > Not sure - any shell expert here? However `*` is usable for example in nmap so I didn't expect any issues and it sounds the most user-friendly for me.
> 
> If you want, we can change to `x`.
> 
> Additionaly, the formatting character shouldn't be hardcoded, maybe some #define is better?

Yes, some define would be good.

AFAIK the shell expands the asterisk e.g. to a filename when it is placed in front or tail:
`AABB**DD` should be ok, but `AABBCCD*` would not be. Therefore `x` or maybe `_` seems safe.

cangen.c
majkk1: MODE_FORMAT changed to MODE_RANDOM_FIX.

cangen.c
majkk1: `*` changed to 'x', define is used

hartkopp:
That's good!

Can you please squash the 4 patches to one patch (with an updated commit message regarding the 'x')?
Thanks!

majkk1:
> That's good!
> 
> Can you please squash the 4 patches to one patch (with an updated commit message regarding the 'x')? Thanks!

Done.

cangen.c
```diff
@@ -398,6 +407,32 @@ enum {
...
+ * Search for CHAR_RANDOM in dataoptarg, save its position, replace it with 0.
+ * Return 1 if at least one CHAR_RANDOM found.
+ */
+static int parse_dataoptarg(char *dataoptarg, unsigned char *rand_position)
+{
+	int mode_format_selected = MODE_FIX;
+	int arglen = strlen(dataoptarg);
+
+	/* Mark nibbles with * as fuzzable */
+	for (int i=0; i < CANFD_MAX_DLEN && i < arglen/2; i++) {
```
marckleinebudde, line 420: Nitpick: spaces around `=` and `/`

cangen.c
```diff
@@ -398,6 +407,32 @@ enum {
...
+		if (optarg[2*i] == CHAR_RANDOM) {
```
marckleinebudde, line 421: same here for the `*`

cangen.c
```diff
@@ -398,6 +407,32 @@ enum {
...
+			optarg[2*i] = '0';
```
marckleinebudde, line 422: dito

cangen.c
```diff
@@ -398,6 +407,32 @@ enum {
...
+			rand_position[i] += NIBBLE_H;
+			mode_format_selected = MODE_RANDOM_FIX;
+		}
+		if (optarg[2*i + 1] == CHAR_RANDOM) {
```
marckleinebudde, line 426: dito

cangen.c
```diff
@@ -398,6 +407,32 @@ enum {
...
+			optarg[2*i + 1] = '0';
```
marckleinebudde, line 427: dito

marckleinebudde (changes requested):

cangen.c
hartkopp: Yes you are right. Missed that entirely.

cangen.c
marckleinebudde: The exact behavior of `*` depends on the shell (and probably configuration). For example on `zsh`:
```shell
echo AABB**DD 
zsh: no matches found: AABB**DD
```

on `bash`:

```shell
echo AABB**DD
AABB**DD
```

If the file exists, both shell will expand the `*`:

```shell
touch AABB--DD
echo AABB**DD
AABB--DD
```

hartkopp:
Committed the patch after applying the style fixes and some more use of CHAR_RANDOM in help texts.

Thanks all!

PR #422, opened 2023-04-06
canlogserver: Update the description

### LINKED ISSUE #421: canlogserver: does it log CAN frame from a remote?
README.md states that
> canlogserver : log CAN frames from a remote/local host

But does it log frames from a remote?  I've been reading the source code but it only `write()`s to `accsocket`, which is the return value of `accept()`.

It does of cause call `socket(PF_CAN, ...)` and `recvfrom()` on the interfaces given.  But does it qualified as a remote? What am I missing?

yashi:
Update the description of canlogserver in README.md and help message.

This closes #421.

### DIFF
--- a/canlogserver.c
+++ b/canlogserver.c
@@ -88,6 +88,7 @@ static volatile sig_atomic_t signal_num;
 
 void print_usage(char *prg)
 {
+	fprintf(stderr, "%s - log CAN frames and serves them.\n", prg);
 	fprintf(stderr, "\nUsage: %s [options] <CAN interface>+\n", prg);
 	fprintf(stderr, "  (use CTRL-C to terminate %s)\n\n", prg);
 	fprintf(stderr, "Options:\n");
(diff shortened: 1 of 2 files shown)


canlogserver.c
```diff
@@ -88,6 +88,7 @@ static volatile sig_atomic_t signal_num;
 
 void print_usage(char *prg)
 {
+	fprintf(stderr, "%s - log CAN frames and serves via TCP.\n", prg);
```
hartkopp, line 91: serves?
Would "serves them" be better here?

canlogserver.c
```diff
@@ -102,7 +103,12 @@ void print_usage(char *prg)
 	fprintf(stderr, "\n");
 	fprintf(stderr, "When using more than one CAN interface the options\n");
 	fprintf(stderr, "m/v/i/e have comma separated values e.g. '-m 0,7FF,0'\n");
-	fprintf(stderr, "\nUse interface name '%s' to receive from all CAN interfaces.\n\n", ANYDEV);
+	fprintf(stderr, "\nUse interface name '%s' to receive from all CAN interfaces.\n", ANYDEV);
+	fprintf(stderr, "\n");
+	fprintf(stderr, "After running canlogserver, connect to it via TCP to get logged data\n");
+	fprintf(stderr, "ex)\n");
```
hartkopp, line 109: is "ex)" a common pattern?

what about "e.g. with 'nc localhost %d'\n", DEFPORT);

canlogserver.c
yashi: That's gonna be

```
$ canlogserver -h
canlogserver - log CAN frames and serves them.
   :
```
If you like it, I'm all for it.

But would it be better to read:
```
$ canlogserver -h
canlogserver - logs local CAN frames and serves to clients.
   :
```

hartkopp:
Thanks @yashi  !

PR #423, opened 2023-04-06
canlogserver: Fix termination after signal

yashi:
With f45de1b782c083, we've removed unsafe function exit(3) from the signal handler.  But that made canlogserver's accept(3)ing process not exit; It unconditionally keeps running with EINTR.

Check the variable "running" when signaled and exit with an appropriate error code.

### DIFF
--- a/canlogserver.c
+++ b/canlogserver.c
@@ -306,6 +306,15 @@ int main(int argc, char **argv)
 			perror("accept");
 			exit(1);
 		}
+		else if ((errno == EINTR) && !running) {
+			close(socki);
+			if (signal_num) {
+				return 128 + signal_num;
+			}
+			else {
+				return 1;
+			}
+		}
 	}
 
 	for (i=0; i<currmax; i++) {


hartkopp:
this commit should be "canlogserver: fix termination after signal"

Because before that there was a proper exit of the application.
Also please add a "Fixes:" tag into the commit message.

yashi:
@hartkopp ping

canlogserver.c
```diff
@@ -306,6 +306,15 @@ int main(int argc, char **argv)
 			perror("accept");
 			exit(1);
 		}
+		else if ((errno == EINTR) && !running) {
```
marckleinebudde, line 309:
```suggestion
		} else if ((errno == EINTR) && !running) {
```

canlogserver.c
```diff
@@ -306,6 +306,15 @@ int main(int argc, char **argv)
...
+			close(socki);
+			if (signal_num) {
+				return 128 + signal_num;
+			}
+			else {
+				return 1;
+			}
```
marckleinebudde, line 316:
```suggestion
			if (signal_num)
				return 128 + signal_num;
			else
				return 1;
```

marckleinebudde (changes requested):
Please use the kernel coding style.

canlogserver.c
marckleinebudde, line 309:
```suggestion
		} else if ((errno == EINTR) && !running) {
```

canlogserver.c
marckleinebudde, line 316:
```suggestion
			if (signal_num)
				return 128 + signal_num;
			else
				return 1;
```

marckleinebudde (changes requested):

PR #439, opened 2023-07-06
cansequence improvements

marckleinebudde:
- add C99 initializers to clear memory
- add support for "any" interface


marckleinebudde:
Thanks!

PR #473, opened 2023-11-27
timestamp formatting: always use 64-bit for timesstamp formatting.

tisyang:
Using C99 `unsigned long long` to format `struct timeval`'s `tv_sec` and `tv_usec`, fix incorrect print under some 32bit platform which using time64.

To fix issue #469. It may looks ugly, but straightforward.

### DIFF
--- a/asc2log.c
+++ b/asc2log.c
@@ -73,7 +73,7 @@ void print_usage(char *prg)
 
 void prframe(FILE *file, struct timeval *tv, int dev, struct canfd_frame *cf, unsigned int max_dlen, char *extra_info) {
 
-	fprintf(file, "(%lu.%06lu) ", tv->tv_sec, tv->tv_usec);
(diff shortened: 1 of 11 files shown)


marckleinebudde:
```
asc2log.c: In function ‘main’:
asc2log.c:438:31: warning: variable ‘tmp_tv’ set but not used [-Wunused-but-set-variable]
  438 |         static struct timeval tmp_tv; /* tmp frame timestamp from ASC file */
      |                               ^~~~~~
```

tisyang:
> ```
> asc2log.c: In function ‘main’:
> asc2log.c:438:31: warning: variable ‘tmp_tv’ set but not used [-Wunused-but-set-variable]
>   438 |         static struct timeval tmp_tv; /* tmp frame timestamp from ASC file */
>       |                               ^~~~~~
> ```

It seems that `tmp_tv` is only used for input check in code, set but not used. I will check again tomorrow and make some change to avoid warning.

tisyang:
@marckleinebudde source code has been updated to fix compile warning `[-Wunused-but-set-variable]` about `tmp_tv`.

PR #491, opened 2024-01-17
canfdtest: fix detection of own frames

marckleinebudde:
Use the correct flag MSG_CONFIRM to detect frames generated on this socket.

Fixes: c35ed80b3d25 ("canfdtest: use struct msghdr::msg_flags to detect own frames")

### DIFF
--- a/canfdtest.c
+++ b/canfdtest.c
@@ -391,7 +391,7 @@ static int can_echo_gen(void)
 				print_frame(rx_frame.can_id, rx_frame.data, rx_frame.len, 0);
 
 			/* own frame */
-			if (flags & MSG_DONTROUTE) {
+			if (flags & MSG_CONFIRM) {
 				err = compare_frame(&tx_frames[recv_tx_pos], &rx_frame, 0);
 				recv_tx[recv_tx_pos] = true;
 				recv_tx_pos++;


rene6262:
Hi marckleinebudde,

This might not be the correct place to ask, but I have been following your git to try and get the CandelightFD device working on a Jetson Nano.  Ive flashed the device with the CANFD supported firmware and it works fine on my Linux CPU (kernel 6.5) but on the Jetson Nano (kernel 4.9) the USB device does not support CANFD commands, only basic CAN.  Is there support on this topic or would you know the solution to this problem?  It looks as if the gs_usb or /net/usb/can drivers need to be updated.

marckleinebudde:
> It looks as if the gs_usb or /net/usb/can drivers need to be updated.

Exactly the kernel you are using is too old,m the `gs_usb` driver doesn't support CAN-FD, yet. If it was not for the Nvidia, you should use the latest kernel. However this external kernel module might help you:
https://github.com/marckleinebudde/gs_usb-external

rene6262:
Thank you marckleinbudde,

I've been trying to make the gs_usb-external but it terminates with this error;
`
root@nffd-1237:/home/nvidia/repos/gs_usb-external# make
make -C /lib/modules/4.9.201-hawkware/build M=/home/nvidia/repos/gs_usb-external modules
make[1]: Entering directory '/usr/src/linux-headers-4.9.201-tegra-ubuntu18.04_aarch64/kernel-4.9'
  CC [M]  /home/nvidia/repos/gs_usb-external/drivers/net/can/usb/gs_usb-core.o
/home/nvidia/repos/gs_usb-external/drivers/net/can/usb/gs_usb-core.c:20:10: fatal error: linux/units.h: No such file or directory
 #include <linux/units.h>
          ^~~~~~~~~~~~~~~
compilation terminated.
`
Following the root of the error, I am looking to get a apt-install linux-headers package, specifically one where units.h will be there but unfortunately I have not had any luck.   

Have you run into this sort of error before? Do you know if I can build this driver on another system and move it over to the Jetson Nano after?

Greatly Appreciated.

marckleinebudde:
> Following the root of the error, I am looking to get a apt-install linux-headers package, specifically one where units.h will be there but unfortunately I have not had any luck.

Your kernel is too old.
 
> Have you run into this sort of error before? Do you know if I can build this driver on another system and move it over to the Jetson Nano after?

No, your kernel is too old.

I've updated https://github.com/marckleinebudde/gs_usb-external, please try again.

PR #496, opened 2024-02-13
isobusfs_cmn: isobusfs_get_timeout_ms(): fix print on 32 bit archs

marckleinebudde:
Fixes the following warning:

```
isobusfs/isobusfs_cmn.c: In function 'isobusfs_get_timeout_ms': | isobusfs/isobusfs_cmn.c:140:51: warning: format '%ld' expects argument of type 'long int', but argument 2 has type 'int64_t' {aka 'long long int'} [-Wformat=]
  140 |                         warn("timeout too long: %ld ms", time_diff);
      |                                                 ~~^      ~~~~~~~~~
      |                                                   |      |
      |                                                   |      int64_t {aka long long int}
      |                                                   long int
      |                                                 %lld
```

### DIFF
--- a/isobusfs/isobusfs_cmn.c
+++ b/isobusfs/isobusfs_cmn.c
@@ -4,6 +4,7 @@
 #include <ctype.h>
 #include <err.h>
 #include <errno.h>
+#include <inttypes.h>
 #include <limits.h>
 #include <stdarg.h>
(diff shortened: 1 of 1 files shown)


olerem:
Looks good to me. Thank you!

PR #498, opened 2024-02-17
Fix compilation on Android

pschichtel:
This fixes a compile error and a warning on Android 13 in a termux environment.

### DIFF
--- a/lib.c
+++ b/lib.c
@@ -45,6 +45,7 @@
 #include <stdint.h>
 #include <stdio.h>
 #include <string.h>
+#include <time.h>
 
 #include <linux/can.h>
 #include <linux/can/error.h>


bcmserver.c
```diff
@@ -156,8 +156,8 @@ int main(void)
 	char rxmsg[50];
 
 	struct {
-		struct bcm_msg_head msg_head;
 		struct can_frame frame;
+		struct bcm_msg_head msg_head;
```
hartkopp, line 160: No! This proposed change is definitely wrong!

```
struct bcm_msg_head {
        __u32 opcode;
        __u32 flags;
        __u32 count;
        struct bcm_timeval ival1, ival2;
        canid_t can_id;
        __u32 nframes;
        struct can_frame frames[];
};
```

The bcm_msg_head is first - followed by a number of CAN frames.

pschichtel: Yeah i already figured this might not make sense, I'll drop that commit. I shouldn't be fixing warnings I have no clue about :D

pschichtel: Dropped it.

PR #502, opened 2024-02-20
isobus: isobusfs_cmn_va.h: remove unused header

marckleinebudde:


marckleinebudde:
IMHO we can just keep it. Just wanted to make sure that it's not some old code.

olerem:
This can be seen as public documentation of the protocol

PR #506, opened 2024-03-17
mcp251xfd-regmap: fix compiler warning

Chaitanya84:
There is no action or impact of the code change just to remove compiler warning during  _make all_

### DIFF
--- a/cangen.c
+++ b/cangen.c
@@ -838,7 +838,10 @@ int main(int argc, char **argv)
 		return 1;
 
 	while (running) {
+		/* clear values but preserve cu.fd.len */
 		cu.fd.flags = 0;
+		cu.fd.__res0 = 0;
+		cu.fd.__res1 = 0;
 
 		if (count && (--count == 0))
 			running = 0;


marckleinebudde:
Which compiler warning are you seeing? Which compiler are you using?

marckleinebudde:
You can use `(void)` in front of `scanf()` to suppress the compiler warning. See #508.

mcp251xfd/mcp251xfd-regmap.c
```diff
@@ -35,7 +35,11 @@ do_mcp251xfd_regmap_read(struct mcp251xfd_priv *priv,
 
 	while ((ret = fscanf(reg_file, "%hx: %x\n", &reg, &val)) != EOF) {
 		if (ret != 2) {
-			fscanf(reg_file, "%*[^\n]\n");
+			//added to remove compile time warning 
+			if(fscanf(reg_file, "%*[^\n]\n"))
+			{
+			asm("nop");
+			}
```
marckleinebudde, line 42: just use `(void)fscanf(reg_file, "%*[^\n]\n");` here

marckleinebudde, line 42: This should fix it:

```suggestion
			int tmp;

			tmp = fscanf(reg_file, "%*[^\n]\n");
			(void)tmp;
```

Issue #516, opened 2024-05-06
bcmserver: build issue with AOSP14


gibsson:
The `bcmserver` tool currently fails to build in AOSP14 tree:
```
[ 97% 158/162] target  C: bcmserver <= external/can-utils/bcmserver.c
FAILED: out/target/product/nitrogen8mp/obj/EXECUTABLES/bcmserver_intermediates/bcmserver.o
/bin/bash -c "PWD=/proc/self/cwd  prebuilts/clang/host/linux-x86/clang-r487747c/bin/clang -I external/can-utils/include -I external/can-utils -I out/target/product/nitroge
n8mp/obj/EXECUTABLES/bcmserver_intermediates -I out/target/product/nitrogen8mp/gen/EXECUTABLES/bcmserver_intermediates -isystem device/boundary/common/kernel-headers -isys
tem device/boundary/common/kernel-headers -Iexternal/libcxx/include -Iexternal/libcxxabi/include -isystem out/soong/.intermediates/bionic/libc/libc/android_vendor.34_arm64
_armv8-a_cortex-a53_shared/gen/include -isystem bionic/libc/kernel/uapi/asm-arm64 -isystem bionic/libc/kernel/uapi -isystem bionic/libc/kernel/android/scsi -isystem bionic
/libc/kernel/android/uapi   -Iexternal/libcxxabi/include    -isystem out/target/product/nitrogen8mp/obj/include -c  -Werror=implicit-function-declaration -DANDROID -fmessa
ge-length=0 -W -Wall -Wno-unused -Winit-self -Wpointer-arith -Wunreachable-code-loop-increment -no-canonical-prefixes -DNDEBUG -UDEBUG -fno-exceptions -Wno-multichar -O2 -
g -fdebug-default-version=5 -fno-strict-aliasing -Werror=date-time -Werror=pragma-pack -Werror=pragma-pack-suspicious-include -Werror=string-plus-int -Werror=unreachable-c
ode-loop-increment -Wno-error=deprecated-declarations -D__compiler_offsetof=__builtin_offsetof -faddrsig -fcommon -Werror=int-conversion -Wno-reserved-id-macro -fcolor-dia
gnostics -Wno-sign-compare -Wno-inconsistent-missing-override -Wno-c99-designator -Wno-gnu-designator -Wno-gnu-folding-constant -Wunguarded-availability -D__ANDROID_UNAVAI
[...]

PR #517, opened 2024-05-06
 bcmserver: use pragma to silence gnu-variable-sized-type-not-at-end warning

marckleinebudde:
closes: https://github.com/linux-can/can-utils/issues/516

### DIFF
--- a/bcmserver.c
+++ b/bcmserver.c
@@ -155,10 +155,13 @@ int main(void)
 	char format[FORMATSZ];
 	char rxmsg[50];
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wgnu-variable-sized-type-not-at-end"
 	struct {
 		struct bcm_msg_head msg_head;
 		struct can_frame frame;
 	} msg;
+#pragma GCC diagnostic pop
 
 	if (snprintf(format, FORMATSZ, "< %%%ds %%c %%lu %%lu %%x %%hhu "
 		     "%%hhx %%hhx %%hhx %%hhx %%hhx %%hhx "
--- a/bcmserver.c
+++ b/bcmserver.c
@@ -194,7 +194,7 @@ int main(void)
 		exit(1);
 	}
 
-	while (1) { 
+	while (1) {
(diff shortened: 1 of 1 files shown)


hartkopp:
OMG! This looks like a really weird hack to silence the (false?) warning!!

So why is `uint8_t _msg_head[sizeof(struct bcm_msg_head)];` not _**variable sized**_ anymore?

Shouldn't we better disable the GNU extension `-Wgnu-variable-sized-type-not-at-end` instead?

marckleinebudde:
`sizeof(struct bcm_msg_head)` is not variable. The `sizeof()` is constant, i.e. without any `struct can_frame` at the end. Another option is to dynamically allocate it with `struct_size`. It doesn't look that ugly.

marckleinebudde:
What about allocating the memory on the heap instead?

marckleinebudde:
Any objections on the `malloc()` approach?

hartkopp:
Yes. I would like to avoid malloc() in can-utils where possible. From my understanding we provide software for embedded systems - and therefore malloc() should be avoided.
 
My preferred idea would still be to disable the -Wgnu-variable-sized-type-not-at-end warning :-/

Btw. what would you think about this approach?
https://github.com/hartkopp/can-utils/commit/47e238ae88fd43398df4e93f651095e2a8f5dd51

marckleinebudde:
> Btw. what would you think about this approach?
> [hartkopp@47e238a](https://github.com/hartkopp/can-utils/commit/47e238ae88fd43398df4e93f651095e2a8f5dd51)

It doesn't fix the warning.

marckleinebudde:
sigh! - gcc complains about unknown pragma
```
 [ 85%] Building C object CMakeFiles/bcmserver.dir/bcmserver.c.o
/home/bcmserver.c: In function 'main':
/home/bcmserver.c:159:32: warning: unknown option after '#pragma GCC diagnostic' kind [-Wpragmas]
  159 | #pragma GCC diagnostic ignored "-Wgnu-variable-sized-type-not-at-end"
      |                                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

PR #522, opened 2024-05-14
Refactor the help/usage message for canplayer

delta512:
Dear maintainers,

as a side effect of PR #521 development, we noticed that the `canplayer` utility returns with an error value, even when asked for the help message explicitly. We thought it might be nicer to limit the above behavior only to unknown options.

We also changed the option from a question mark `-?` into an `-h`, which is more common among CLI tools.

Hope you'll find these nuances sound and do not hesitate to request changes.

Have a great day again!


marckleinebudde:
Thanks for the contribution. Please add `-h` to `print_usage()`.

delta512:
> Thanks for the contribution. Please add `-h` to `print_usage()`.

Oh, I really missed on this. I amended the previous commit so it's all well packed together.

Hope it looks good know

Issue #526, opened 2024-05-23
Too many build systems


marckleinebudde:
Hey,

we have way too many build systems for this packet. I want to remove at least 1, better 2. I think I'd like to keep cmake, `Android.mk` and remove the others.

Marc

hartkopp:
How many build systems do we really have right now?

I still like `make && make install` - but of course I have cmake on my system too.

marckleinebudde:
Besides Android, we have autotools, cmake and plain old Makefiles.

PR #527, opened 2024-05-23
build-systems: remove autotools and Android.mk, upgrade CI infrastructure

marckleinebudde:
Remove plain old Makefile and autotools support. This clauses less friction when adding new tools.

Link: https://github.com/linux-can/can-utils/issues/526


marckleinebudde:
Our github workflows only runs autotools, I'll convert it to camke.

hartkopp:
IMHO removing autotools and its infrastructure is a good choice.
No need to have autotools and cmake in parallel as these are doing the same.

But I'm not sure if it is a good idea to drop the 'plain' Makefile as it enables (embedded) systems to build and run can-utils without any additional build system.

Instead I would check the possibility to build the Android binaries with cmake too.
There are ongoing discussions which suggest to use cmake in new projects for Android:
https://stackoverflow.com/questions/39589427/difference-between-cmake-and-ndk-build-in-android-studio-project

And then we finally would have a cmake and a plain Makefile support.

marckleinebudde:
> But I'm not sure if it is a good idea to drop the 'plain' Makefile as it enables (embedded) systems to build and run can-utils without any additional build system.

I think it's a very good idea to drop plain Makefile support.

You want to build can-utils on your embedded system? IMHO if you have a compiler, having cmake doesn't make a big difference.

marckleinebudde:
Regarding Android. I'm not invested, don't use it and cannot test it. If someone upgrades the `CMakeLists.txt`, so that we can remove the `Android.mk`, I'm more than happy!

yegorich:
The last patch should be called "build-system: remove Anroid.mk".

marckleinebudde:
Wanted to `commit --amend` into the 2nd patch. fixed. Thanks.

hartkopp:
> As to the plain Makefiles on an embedded system. This is mostly possible on more or less standard distors like Debian/Ubuntu. Builroot, OpenWrt, etc. don't provide a compiler as a target package. So, in most cases you'll have to cross-compile your code.

Ok. I recently got a Ubuntu 14.04 (with a now unsupported cmake 2.8) on my desk and wanted to build the latest can-utils. 
The entire repo is enabled to build the can-utils even on very old Linux installations today. And I really like that feature.

As you now removed 50% of the build systems (especially the ugly ones) this is already a nice cleanup.
The plain old Makefile does not seem to be hard to maintain together with CMakeLists.txt and it is a good 'learning' example.  Therefore I really would like to keep the plain old Makefile together with the cmake build system.

.github/workflows/compile.yml
```diff
@@ -0,0 +1,46 @@
...
+        podman exec -i -u root stable apt-get update
+        podman exec -e DEBIAN_FRONTEND='noninteractive' -i -u root stable apt-get install -qy build-essential cmake ninja-build
+
+    - name: Configure
+      run: |
+        podman exec -i stable cmake -B build
+
+    - name: Build
+      run: |
+        podman exec -i stable make -C build
```
yegorich, line 46: What about using `cmake --build build`? This command was introduced in cmake 3.13 that old enough to satisfy used distro releaes.

marckleinebudde:
FWIW I've kept the plain old `Makefile`. I personally consider it obsolete. I've not added it to the new CI integration.

marckleinebudde:
@yegorich 

> We already have all the [instructions](https://github.com/linux-can/can-utils?tab=readme-ov-file#cmake-project-generator) required for an Android build. Just tried to build against android-ndk-r26d and it works. Hence, we can also remove `Android.mk`.

Can you create a android CI integration?

marckleinebudde:
> I'll look into it. GitHub images seem to also ship NDK

That was quick! Thanks a lot!

Issue #525, opened 2024-05-22
New tools: canerrsim and canerrdump


zeljkoavramovic:
Developing robust CAN software is not easy when you do not have a quick way to generate all kinds of CAN errors. So I created such tool and called it canerrsim. Although candump is able to show CAN errors, I have created canerrdump which is specialized to do just that.

I would like to donate both canerrsim and canerrdump to can-utils project, but before doing that I would like to discuss them here and find out if anything needs to be changed to reach that goal (like different code formatting, or adding features like timestamp to canerrdump).

For now, you can find source code of both tools in the attachments.

[canerrsim.zip](https://github.com/linux-can/can-utils/files/15408116/canerrsim.zip)
[canerrdump.zip](https://github.com/linux-can/can-utils/files/15408118/canerrdump.zip)

marckleinebudde:
Hey @zeljkoavramovic,

thanks for your contribution. `canerrsim` looks interesting, but I think the functionality of `canerrdump` is already covered with `candump`. Compare the output of both tools:

```
$ candump any,0:0,#FFFFFFFF -cexdtd
 (590.634662)  vcan0  TX - -  20000025   [8]  00 40 00 00 00 00 00 00   ERRORFRAME
        tx-timeout
        controller-problem{back-to-error-active}
        no-acknowledgement-on-tx
```

```
$ ./canerrdump vcan0
CAN Sockets Error Messages Dumper
Listening CAN bus vcan0 for errors...
0x025 [8] 00 40 00 00 00 00 00 00  ERR=TxTimeout,NoAck,Ctrl(Active)
```

zeljkoavramovic:
Hello @marckleinebudde,

I thank you for your comments and I am glad you like canerrsim. Please let me know if there are any changes needed before it's acceptance.

It's fine if canerrdump is not accepted, but please let me try to explain why I still think it should be. There are 2 main use cases for canerrdump: troubleshooting CAN errors, and as a checker tool when you develop CAN application that yow want to react properly on some specific CAN errors. In both cases it is beneficial if you can hide regular messages, show all errors, or ignore some specific annoying recurring errors that are not in your current focus. This is where specialized tool like canerrdump shines. For example, using simple switches like IgnoreTxTimeout, IgnoreController and IgnoreBusError is very descriptive and does the job quickly. While it can be done with candump, you would need to browse through many documents and do some bit calculations before figuring out the proper hex mask. What do you think?


Btw, I have just found one issue while testing:

[...]

PR #528, opened 2024-05-23
canerrsim

marckleinebudde:
Add new tool `canerrsim` proposed by @zeljkoavramovic in https://github.com/linux-can/can-utils/issues/525


marckleinebudde:
Intgration in `Android.mk` is missing.

zeljkoavramovic:
> Intgration in `Android.mk` is missing.

If I understood well Android.mk file is deleted and no longer used, so it should be safe to continue with canerrsim integration

yegorich:
@marckleinebudde `#include <ctype.h>` can be removed.

PR #532, opened 2024-05-26
Cmake fixes

yegorich:


marckleinebudde:
My cmake was complaining about the minimum version, too. Thanks.

PR #534, opened 2024-05-27
CMake: resolve linting issues

yegorich:
Improve indentation according to the .editorconfig style.

Write commands in lowercase letters.

Break long lines.

Remove double keywords (PRIVATE).


marckleinebudde:
Quite a spring cleaning, you're doing. :)

yegorich:
If we want to use CMake as the only build system, it has to look perfect :-)

yegorich:
What I would also like to do is to move source files into the `src` folder. This way, the root folder will look much cleaner. But it can be done only after purging the other build systems.

Issue #536, opened 2024-05-28
gcc complains about unknown pragma


marckleinebudde:
sigh! - gcc complains about unknown pragma
```
 [ 85%] Building C object CMakeFiles/bcmserver.dir/bcmserver.c.o
/home/bcmserver.c: In function 'main':
/home/bcmserver.c:159:32: warning: unknown option after '#pragma GCC diagnostic' kind [-Wpragmas]
  159 | #pragma GCC diagnostic ignored "-Wgnu-variable-sized-type-not-at-end"
      |                                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

_Originally posted by @marckleinebudde in https://github.com/linux-can/can-utils/issues/517#issuecomment-2134804757_

PR #537, opened 2024-05-28
bcmserver: use pragma to silence unknown pragma option

marckleinebudde:
Silence the following warning by switching it of via a pragma.

| bcmserver.c: In function 'main':
| bcmserver.c:159:32: warning: unknown option after '#pragma GCC diagnostic' kind [-Wpragmas]
|   159 | #pragma GCC diagnostic ignored "-Wgnu-variable-sized-type-not-at-end"
|       |                                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Closes: https://github.com/linux-can/can-utils/issues/536

### DIFF
--- a/bcmserver.c
+++ b/bcmserver.c
@@ -156,6 +156,7 @@ int main(void)
 	char rxmsg[50];
 
 #pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wpragmas"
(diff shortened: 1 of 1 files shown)


hartkopp:
Thanks!!

PR #543, opened 2024-05-29
.github/workflows/ndk.yml: add workflow for NDK build

yegorich:


marckleinebudde:
Thanks for looking into this. I've added this to https://github.com/linux-can/can-utils/pull/527, switched to `actions/checkout@v4` and tweaked it to look similar to by other workflow.

PR #549, opened 2024-07-16
j1939acd: fix uclibc compile (const char* program_invocation_name)

pseiderer:
- disable command line option '-p' in case of uclibc

Fixes:
```
  j1939acd.c: In function 'main': j1939acd.c:489:38: error: passing argument 1 of 'asprintf' from incompatible pointer type [-Wincompatible-pointer-types]
    489 |                         if (asprintf(&program_invocation_name, "%s.%s",
        |                                      ^~~~~~~~~~~~~~~~~~~~~~~~
        |                                      |
        |                                      const char **
```

### DIFF
--- a/j1939acd.c
+++ b/j1939acd.c
@@ -485,12 +485,16 @@ int main(int argc, char *argv[])
 			s.current_sa = strtoul(optarg, 0, 0);
 			break;
 		case 'p':
-#ifdef _GNU_SOURCE
+#if defined(_GNU_SOURCE) && !defined(__UCLIBC__)
 			if (asprintf(&program_invocation_name, "%s.%s",
 				     program_invocation_short_name, optarg) < 0)
 				err(1, "asprintf(program invocation name)");
 #else
+#if defined(__UCLIBC__)
+			err(0, "option -p disabled with uclibc");
+#else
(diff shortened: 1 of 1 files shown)


marckleinebudde:
What about removing the `-P` to set the `program_invocation_name` functionality from all programs?

olerem:
Hm, I have nothing against it.

pseiderer:
Fine with me too ;-), specially as the following test program
```
define _GNU_SOURCE
#include <stdio.h>
#include <errno.h>
#include <err.h>

int main(int argc, char* argv[]) {
        program_invocation_name = program_invocation_short_name;
        asprintf(&program_invocation_name, "bla");
        asprintf(&program_invocation_short_name, "blub");
        warn("warn-test");
        err(1, "err-test");
        return 0;
}
```
produces the following output:
```
blub: warn-test: Success
blub: err-test: Success
```
Seems asprintf of j1939acd.c overwrites the wrong variable (at least for my local system)...

Will provide an alternative pull-request removing the program_invocation_name overwrite funcionality...

olerem:
Close it in favor of https://github.com/linux-can/can-utils/pull/552

PR #556, opened 2024-08-08
Fix and rework canbusload

hartkopp:
- rework to use a single CAN socket for the received CAN traffic.
This allows to shutdown and restart various CAN interfaces without
terminating the application.
Additionally an auto detection has been implemented with the 'any' CAN
interface. E.g. with any@500000,2000000 CAN interfaces that have not
been defined before are assigned with the given 'any' bitrates when a
CAN frame from this CAN interface has been received.

- show RX/TX direction in bargraph
- fix and improve the bitrate output

### DIFF
--- a/canbusload.c
+++ b/canbusload.c
@@ -70,11 +70,13 @@
 
 #define PERCENTRES 5 /* resolution in percent for bargraph */
 #define NUMBAR (100 / PERCENTRES) /* number of bargraph elements */
+#define BRSTRLEN 20
 
 extern int optind, opterr, optopt;
 
 static struct {
 	char devname[IFNAMSIZ + 1];
+	char bitratestr[BRSTRLEN]; /* 100000/2000000 => 100k/2M */
 	char recv_direction;
 	int ifindex;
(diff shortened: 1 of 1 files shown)


canbusload.c
```diff
@@ -243,14 +246,16 @@ static void printstats(int signo)
...
-
+	int s;
 	int opt;
 	char *ptr, *nptr;
 	struct sockaddr_can addr;
 	struct canfd_frame frame;
 	int nbytes, i;
-	struct ifreq ifr;
+
+	int have_anydev = 0;
```
marckleinebudde, line 287: `bool`?

marckleinebudde (changes requested):
just scrolled over the changes....

canbusload.c
hartkopp: no. This is needed for an offset calculation later.

hartkopp: I currently have a setup with 16 interfaces on my desk - partly with CAN FD.
And it turned out that the canbusload tool needed some love ;-)

PR #571, opened 2025-01-20
cangen, canbusload: new features

marckleinebudde:


marckleinebudde:
Makes sense, please create a new issue.

PR #577, opened 2025-02-10
Update .gitignore

jengelh:


.gitignore
```diff
@@ -40,3 +44,6 @@ tags
 /slcand
 /slcanpty
 /testj1939
+*.so
+*.so.*
+*.a
```
marckleinebudde, line 49: Move these to the block in the beginning

.gitignore
```diff
@@ -3,6 +3,10 @@
 .ccls-cache
 compile_commands.json
 tags
+Makefile
+CMakeCache.txt
+CMakeFiles/
+cmake_install.cmake
```
marckleinebudde, line 9: Please keep this block alphabetically sorted.

marckleinebudde (changes requested):

.gitignore
jengelh: i'll use ascii sorting

yegorich:
Doesn't it make problems to ignore the existing `Makefile`?

marckleinebudde:
Doh! Although I personally want to get rid of handcrafted `Makefile`s, that change is obviously wrong. Please send a fix.

jengelh:
Oh.. I guess this an artifact of cmake (and previously, automake) generating a Makefile and replacing the shipped one, suggesting there was erroneous tracking of it:

```
$ cmake . && make -j32
…
$ git st
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   Makefile
```

yegorich:
Perhaps, we should rewrite `README.md` and suggest using out-of-source builds:

`cmake -B build && cmake --build build`

marckleinebudde:
you should use cmake in a separate build dir

marckleinebudde:
Yeah, the `README.md` just has some `..` at the end of the `cmake` calls, but doesn't mention the `mkdir build; cd build;` you should do before. Using `cmake -B`/`cmake --build` is even better.

hartkopp:
Ooops! Sorry.
I just looked at the reordering that Marc requested :-(

marckleinebudde:
I've missed the `Makefile` change, too.

hartkopp:
Add the fix myself quickly.

PR #592, opened 2025-03-10
Use timespec

RikusW:
Use timespec instead of timeval so timestamp values with nanosecond resolution can be displayed.

### DIFF
--- a/canplayer.c
+++ b/canplayer.c
@@ -452,14 +452,20 @@ int main(int argc, char **argv)
 			return 1;
 		}
 		log_tv.tv_sec = sec;
-		log_tv.tv_usec = usec;
 
 		/*
-		 * ensure the fractions of seconds are 6 decimal places long to catch
+		 * ensure the fractions of seconds are 6 or 9 decimal places long to catch
 		 * 3rd party or handcrafted logfiles that treat the timestamp as float
 		 */
-		if (strchr(buf, ')') - strchr(buf, '.') != 7) {
-			fprintf(stderr, "timestamp format in logfile requires 6 decimal places\n");
+		switch (strchr(buf, ')') - strchr(buf, '.')) {
+		case 7: //6
+			log_tv.tv_usec = usec;
+			break;
+		case 10: //9
+			log_tv.tv_usec = usec / 1000;
+			break;
+		default:
+			fprintf(stderr, "timestamp format in logfile requires 6 or 9 decimal places\n");
 			return 1;
 		}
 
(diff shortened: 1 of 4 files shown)


marckleinebudde:
I think we're not allowed to change the output for the `-l` format. The tool `canplayer` relies on 6 decimal places:

```
candump any -l &
cangen vcan0 -I 2 -Li -Di -p 10 -g 100 -n 10
killall candump
canplayer -I ./foo
```

Otherwise this is a good enhancement for the hardware timestamps.

hartkopp:
Yes. The 6 decimal places are fix in the CAN log file format for 20+ years now. And having the maximum rate of CAN frames on the CAN bus in mind we are already far beyond what is needed as a timestamp precision.

Of course everyone might use a higher precision from timespec inside a SocketCAN application - but we have to stay with 6 decimal places in the log file format.

RikusW:
We are using MCP2518FD with 25ns resolution and are planning to correlate that to PTP time.
I've made the nanosecond format optional with microseconds being default.

marckleinebudde:
@RikusW Can you elaborate on you use case a but more, I have some proof of concept patches that implement a PTP clock for the mcp251xfd driver.

RikusW:
@marckleinebudde We will be using 8 CAN ports with all of them exactly in sync and will be doing the PTP part in userspace for now, I'm working on https://tsn.systems/en/products/basicsolution/tsn-basicsolution-octocan/

hartkopp:
I'm not generally against creating a new log file format flavor that supports a nanosecond resolution with a `-N` option.
But the highest impact to can-utils will be to make the consumers of those new log files , especially `canplayer` and `log2asc` aware of 9 digits and adapt themselves at runtime. These tools need to be converted until `candump` starts to provide such log files.

RikusW:
For now I just need to be able to view nanosecond timestamps, transmit/replay ability could be done at a later time even.
Adding us/ns autodetection should be easy enough, I'll have a look at that a bit later.

hartkopp:
Let's hold this PR in mind until it is completed with the other tools then.

marckleinebudde:
> Looks like libc broke in some test setups...

yes - debian and ubuntu have some problems.

RikusW:
> > Looks like libc broke in some test setups...
> 
> yes - debian and ubuntu have some problems.

At least it is not my patch that is the problem here...

marckleinebudde:
ACK

RikusW:
Any news on the broken distributions ?

log2asc.c
```diff
@@ -407,7 +407,22 @@ int main(int argc, char **argv)
 			}
 		}
 		tv.tv_sec = sec;
-		tv.tv_usec = usec;
+
+		/*
+		 * ensure the fractions of seconds are 6 or 9 decimal places long to catch
+		 * 3rd party or handcrafted logfiles that treat the timestamp as float
```
hartkopp, line 413: I would prefer a comment that talks about nanoseconds instead of float.

canplayer.c
```diff
@@ -541,19 +547,25 @@ int main(int argc, char **argv)
 					break;
 				}
 
-				if (sscanf(buf, "(%llu.%llu) %s %s", &sec, &usec, device, afrbuf) != 4) {
+				if (sscanf(buf, "(%llu.%llu) %21s %6299s", &sec, &usec, device, afrbuf) != 4) {
```
hartkopp, line 550: limiting the string reading is a good idea but is not in the scope of this patch.
Please create a separate patch with a separate commit message for it.

log2asc.c
RikusW: I kept this comment the same as in canplayer.c except for adding "or 9"

canplayer.c
RikusW: Removed this, I can squash, or it can be squashed when merging ?

RikusW:
Looks like the 2 failing checks is the distribution build system failing and unrelated to this patch ?

PR #594, opened 2025-04-22
Add canperformance.c and modify Makefile for CAN performance testing.

fafactx:
### What's Done
1. Added a new file named canperformance.c to the project.
2. Modified the Makefile to support the compilation of canperformance.c.

### Core Idea
The addition of canperformance.c aims to meet CAN performance testing needs, compensating for cangen's deficiencies.

### Key Features
- Support for standard and extended frames
- Support for file sending and receiving
- Real - time frame rate display
- MD5 verification for CAN ID, Data, or files

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,6 +53,7 @@ set(PROGRAMS_CANLIB
   canbusload
   candump
   cangen
+  canperformance
   canplayer
   cansend
   cansequence
--- a/Makefile
+++ b/Makefile
@@ -123,7 +123,8 @@ PROGRAMS := \
 	log2asc \
 	log2long \
 	mcp251xfd-dump \
-	slcanpty
+	slcanpty \
+	canperformance
 
 ifeq ($(HAVE_FORK),1)
 PROGRAMS += \
@@ -227,4 +228,7 @@ can-calc-bit-timing: calc-bit-timing/can-calc-bit-timing.o
 	$(CC) $(LDFLAGS) $^ $(LDLIBS) -o $@
 
 mcp251xfd-dump: mcp251xfd/mcp251xfd-dev-coredump.o mcp251xfd/mcp251xfd-dump.o mcp251xfd/mcp251xfd-main.o mcp251xfd/mcp251xfd-regmap.o
+	$(CC) $(LDFLAGS) $^ $(LDLIBS) -o $@
+
+canperformance: canperformance.o
 	$(CC) $(LDFLAGS) $^ $(LDLIBS) -o $@
--- a/canperformance.c
+++ b/canperformance.c
@@ -1630,7 +1630,6 @@
      int expected_frame_count = 0;
      int frames_received = 0;
      uint8_t *file_buffer = NULL;
-     long bytes_written = 0;
 
      /* Variables for progress bar */
      int progress_bar_width = 50;
@@ -1850,7 +1849,6 @@
              if (bytes_to_write > 0) {
                  memcpy(file_buffer + offset, frame.data,
                         bytes_to_write);
-                 bytes_written += bytes_to_write;
              }
 
              frames_received++;
--- a/canperformance.c
+++ b/canperformance.c
@@ -1,0 +1,2273 @@
+/*
+ * Copyright (c) 2023-2025
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/***********************************************************************************************************************
+ * canperformance.c - SocketCAN performance testing utility
+ *
+ * Author: Ken Li (ken.li@nxp.com) - NXP Semiconductors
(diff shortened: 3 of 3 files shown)


hartkopp:
Hi @fafactx ,

many thanks for your contibution.

Btw. are you aware of the tools `cangen` and `canbusload` ? Or the `canfdtest` tool? (with all their options?)

There is a Unix tool design concept with the idea of "**one** tool doing **one** job well" 

Your new tool is a mainly combination of  `cangen` and `canbusload` while those tools already support CAN FD and `cangen` also supports CAN XL.

Do you feel something missing in the existing features of the existing tools for your work?

fafactx:
Hi @hartkopp ，
Agreed, cangen and similar tools are great, and I rely on them in my projects.
This PR came about because I needed to measure CAN FPS and detect bit errors. While CAN's physical layer doesn't guarantee error - free operation, I specifically wanted to test for zero bit errors at maximum FPS. For 1M standard CAN frames, that's around 7500 FPS.
None of the existing tools could do this. As you said, one tool should do one job well, and this tool focuses on CAN performance testing. Future plans include adding FD and XL support.

marckleinebudde:
Have you ever seen a CAN frame where you have detected an error using the MD5? CAN-CC uses a 15 bit CRC, which can detect single and double bit errors.

There's another strategy to detect errors: by sending predicable data. That's what `canfdtest` does. Though it's full duplex and doesn't do full speed to interlace RX and TX messages.

marckleinebudde:
I've added the tool to cmake. https://github.com/fafactx/can-utils/commit/97e5e0ddc221b94f1636cc8aa33410929ea0e059

marckleinebudde:
To measure the CAN FPS you can use `canbusload`.

hartkopp:
Exactly! The FPS value is one of the most essential information items in `canbusload` which is printed every second with additional other information.
 
canbusload2

fafactx:
Hi @hartkopp @marckleinebudde,
You both mentioned canbusload. I was using it in the early stage, and it was really great. However, in the later stage, since I needed to conduct a stress test and ensure that the FPS records were completed at a speed close to the CAN line speed, it couldn't meet my requirements.
Of course, in the early stage, I used a shell script to complete the sending and receiving verification. But after all, it wasn't very elegant. As I had plenty of time later on, I wrote this tool to make it convenient for the stress tests of future projects.
Thank you again. If you think this tool does not meet the merging requirements, we can close this PR. Keeping it here may provide some ideas and inspiration for developers in need.

marckleinebudde:
Have you ever seen a CAN frame where you have detected an error using the MD5?

fafactx:
Hi @marckleinebudde 
Yes, I've indeed encountered situations where errors in CAN frames were detected using MD5. However, it's evident that these errors weren't caused by issues at the physical layer. My test model involves data transmission from a Linux system to an MCU via the SPI interface, and then from the MCU to the CAN bus.

After analysis, there are mainly two scenarios that lead to MD5 verification errors. One is when the MCU fails to process the CAN frames sent from the Linux system via SPI in a timely manner. Since we've set a very high transfer rate for the SPI interface, the MD5 checksum may not match. The other scenario is that when we introduce abnormal interference on the CAN bus, it can also trigger MD5 errors.

The purpose of conducting such tests is primarily to verify and tune the maximum CAN forwarding capacity of the MCU, so as to optimize the system performance more effectively.

marckleinebudde:
> One is when the MCU fails to process the CAN frames sent from the Linux system via SPI in a timely manner. Since we've set a very high transfer rate for the SPI interface, the MD5 checksum may not match.

Seems your CAN driver doesn't implement proper flow control so that CAN frames get lost somewhere between the kernel's CAN driver and the MCU. Using `canfdtest` and `cangen` + `cansequence` would be my tools of choice here.

* `canfdtest` doesn't work at line rate, but it's a good idea to make the data rate configurable
* `cangen` + `cansequence` is one way communication only, but at line rate

Both tools detect missing and/or re-ordered frames.

> The other scenario is that when we introduce abnormal interference on the CAN bus, it can also trigger MD5 errors.

Ok, the interference messes up CAN frames. The question is, do you receive CAN frames with garbage content in the application or are there frames missing, as CAN frames with broken CRC are discarded. On the other hand, a CAN frame with broken CRC should not be ACKed and thus re-transmitted by the sender, unless you enable one-shot-transmission.

fafactx:
For case 1, you're absolutely right. Flow control is indeed necessary. Currently, it hasn't been incorporated into the first version of the driver. Once the adaptation of the MCU firmware is completed, we will improve the flow control for the downstream transmission.
For case 2, we haven't conducted in-depth research on it yet. In the following stage, we will observe it in combination with the CAN analysis instrument. I will update any new findings here. I need to confirm with the MCU about the configuration of the CAN controller.

PR #596, opened 2025-05-21
Allow more fine-grained control over CMake targets

scootermon:
Currently the CMake build is either all or nothing. You have explicit control over which targets are built using `cmake --build --target <xyz>`, but this doesn't apply to install step.
Running `cmake --install` will currently always install all available targets.

This PR adds the same groups that are present in the README to the CMake build, allowing you to control them individually. By default everything is turned on, which means there should be no differences compared to before.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,7 +87,7 @@ if(ENABLE_BASIC_TOOLS)
 endif()
 
 if(ENABLE_IP_SERVER AND HAVE_FORK)
-  list(APPEND PROGRAMS        bcmserver)
+  list(APPEND PROGRAMS bcmserver)
   list(APPEND PROGRAMS_CANLIB canlogserver)
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,6 +21,25 @@ option(ENABLE_WERROR "Treat all compiler warnings as errors" OFF)
 option(ENABLE_GPS "Enable GPS support" OFF)
 option(BUILD_SHARED_LIBS "Build shared libraries" ON)
 
+# Options for controlling targets. These groups correspond to the README.
+option(ENABLE_BASIC_TOOLS "Build basic tools" ON)
+option(ENABLE_IP_SERVER   "Build tools providing CAN access via IP sockets" ON)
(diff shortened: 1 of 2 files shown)


marckleinebudde:
Please fix the NDK build https://github.com/linux-can/can-utils/actions/runs/15161415721/job/42631214602?pr=596

CMakeLists.txt
```diff
@@ -48,108 +60,167 @@ include_directories(./include)
...
   )
+endif()
+
+if(can-calc-bit-timing IN_LIST PROGRAMS)
+  list(REMOVE_ITEM PROGRAMS can-calc-bit-timing)
+  add_executable(can-calc-bit-timing
+    calc-bit-timing/can-calc-bit-timing.c
+  )
+  install(TARGETS can-calc-bit-timing DESTINATION ${CMAKE_INSTALL_BINDIR})
+endif()
```
marckleinebudde, line 209: Can you move the:

```
  list(REMOVE_ITEM PROGRAMS can-calc-bit-timing)
  add_executable(can-calc-bit-timing
    calc-bit-timing/can-calc-bit-timing.c
  )
  install(TARGETS can-calc-bit-timing DESTINATION ${CMAKE_INSTALL_BINDIR})
```

to the place where `can-calc-bit-timing` is added to the `PROGRAMS` in the first place? Same for the other `list(REMOVE_ITEM ...)`?

marckleinebudde (changes requested):

CMakeLists.txt
scootermon: Of course. I was hoping to create a bit of separation between the "program selection" and the actual build part, but I can see where you're coming from.

CMakeLists.txt
```diff
@@ -48,103 +67,111 @@ include_directories(./include)
...
+    canplayer
+    cansend
+    cangen
+    cansequence
+  )
+endif()
 
-if(HAVE_FORK)
+if(ENABLE_IP_SERVER AND HAVE_FORK)
+  list(APPEND PROGRAMS        bcmserver)
```
marckleinebudde, line 90:
```suggestion
  list(APPEND PROGRAMS bcmserver)
```

I'll fix that.

marckleinebudde (approved):

PR #600, opened 2025-09-08
cangen: make FD/XL content configurable in mixed mode

hartkopp:
Suggested-by: Vincent Mailhol <mailhol@kernel.org>


vincent-mailhol:
```c
	if (mix && (!canfd_en && !canxl_en)) {
		printf("-f and/or -X is required for -m (mixed mode)!\n");
		return 1;
	}
```

This is not what I intended in my initial patch. The goal was to have the cangen
automagically handle the logic for me so that I do not have to think about what
flag to pass and this regardless of the interface capabilities.

This actually goes in the opposite direction as I would now not be able to use
-m alone.

When testing, I use the cangen -m a lot an I always have this little frustration
when getting the error message when the interface only support Classical CAN.

PR #601, opened 2025-09-08
cangen: auto enable FD/XL content in mixed mode

hartkopp:
Automatically create FD/XL content in mixed mode when the CAN interface is capable to deal with it.

Suggested-by: Vincent Mailhol <mailhol@kernel.org>


marckleinebudde:
Cc: @vincent-mailhol

PR #619, opened 2026-03-04
Move 'gap' of canplayer between frames not logs.

JamesB192:
For when -t was too fast, and sans -t, takes too long.

### DIFF
--- a/canplayer.c
+++ b/canplayer.c
@@ -580,10 +580,10 @@ int main(int argc, char **argv)
 					last_log_tv = log_tv;
 				}
 
-			} /* while frames_to_send ... */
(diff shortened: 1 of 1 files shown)


hartkopp:
I get your point but the patch is wrong IMO.

If you have a bunch of frames that are so tight together that they are needed to be pushed out until the next waiting happens, you simply have to do so. Otherwise you will never get the correct timely behaviour.

This gap setting is more like a "how often the loop is running" than describing the gap between two CAN frames.

And inside such a (1ms) loop multiple CAN frames may be needed to be processed when their timestamps fit into this 1ms section.

PR #628, opened 2026-05-06
canlogserver: fix infinite loops during signal handling

olerem:
Fix infinite loops that prevent graceful termination when the server receives SIGINT or SIGTERM signals. Without this fix, Ctrl-C and kill commands are ignored, making it impossible to stop the server cleanly.

Two scenarios cause the infinite loops:

1) The bind() retry loop: When the port is busy, the loop retries
   indefinitely without checking the running flag set by the signal
   handler.

2) The accept() loop: The loop is unconditional, so when accept() is
   interrupted by a signal and returns EINTR, the loop immediately
   restarts, ignoring the shutdown request.

### DIFF
--- a/canlogserver.c
+++ b/canlogserver.c
@@ -177,7 +177,8 @@ int main(int argc, char **argv)
 	sigset_t sigset;
 	fd_set rdfs;
 	int s[MAXDEV];
-	int socki, accsocket;
+	int socki;
+	int accsocket = -1;
 	canid_t mask[MAXDEV] = {0};
 	canid_t value[MAXDEV] = {0};
 	int inv_filter[MAXDEV] = {0};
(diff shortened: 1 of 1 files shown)


canlogserver.c
```diff
@@ -295,25 +296,38 @@ int main(int argc, char **argv)
 		nanosleep(&f, NULL);
 	}
 
+	if (!running) {
```
hartkopp, line 303: The while() statement checks for running to be 1 so this check seems to be wrong inside the function.
I would suggest some comment here that running might be set to 0 due to an external signal.

canlogserver.c
```diff
@@ -295,25 +296,38 @@ int main(int argc, char **argv)
...
 			close(accsocket);
+			accsocket = -1;
 		}
 		else if (errno != EINTR) {
 			perror("accept");
 			exit(1);
 		}
 	}
 
+	if (!running) {
```
hartkopp, line 332: same here

hartkopp:
Btw. thanks for the fix!

hartkopp:
Thanks @olerem

PR #633, opened 2026-09-02
lib: bound parse_canframe input length

### LINKED ISSUE #632: lib: parse_canframe: strlen OOB read when input is not null-terminated
## Summary

`parse_canframe()` in `lib.c` calls `strlen(cs)` as its first operation
without verifying that the input buffer is null-terminated. Passing a
buffer without a null terminator causes an out-of-bounds read in `strlen`,
detected by AddressSanitizer as a heap-buffer-overflow.

## Affected code

`lib.c`, line 168:
```c
int parse_canframe(char *cs, union cfu *cu)
{
    int len;
    len = strlen(cs);   // ← OOB read if cs has no null terminator
    ...
}
```

## Call sites at risk

`parse_canframe` is called with file-read buffers (not just CLI argv),
which may not be null-terminated if the input is malformed or crafted:

- `canplayer.c:509` — `parse_canframe(afrbuf, &cu)`
- `log2asc.c:449`   — `parse_canframe(afrbuf, &cu)`
- `log2long.c:84`   — `parse_canframe(afrbuf, &cu)`

## Reproduction (minimal)

```c
#include "lib.h"
#include <string.h>

int main(void) {
    union cfu cu;
    char input[1] = {0x0a};   // 1 byte, no null terminator
    parse_canframe(input, &cu);
    return 0;
}
```

Compile and run:
```bash
clang -g -fsanitize=address \
  -DCANXL_VCID_MASK=0x00FF0000 -DCANXL_VCID_OFFSET=16 \
  -o repro repro.c lib.c
./repro
```

## ASAN output

​```
==ERROR: AddressSanitizer: stack-buffer-overflow
READ of size 2 at ... thread T0
    #0 in strlen
    #1 in parse_canframe lib.c:168
    #2 in main repro.c:7
SUMMARY: AddressSanitizer: stack-buffer-overflow in strlen
​```

## Suggested fix

Add a `size_t len` parameter so callers can pass the known buffer length,
and replace `strlen` with `strnlen` or an explicit length check:

```c
// Option A — add length parameter
int parse_canframe_n(char *cs, size_t cs_len, union cfu *cu);

// Option B — guard strlen with a maximum length
size_t len = strnlen(cs, MAX_CANFRAME_STR_LEN);
if (len == MAX_CANFRAME_STR_LEN) return 0;  // reject unterminated input
```

## Discovery

Found using an LLM-assisted fuzzing pipeline (LogosFuzz) with
libFuzzer + AddressSanitizer. Triggered with a 1-byte input (`0x0a`)
in under 30 seconds from an empty corpus.

seldatrgt35:
## Summary

Pass the available input buffer size to `parse_canframe()` and reject
input that has no null terminator within that bound.

Update all four callers to provide the available buffer size. This
prevents the unbounded `strlen()` read reported in #632 while preserving
the existing parsing behavior for valid CAN CC, CAN FD, CAN XL, and RTR
frame strings.

## Validation

- Built `cansend`, `canplayer`, `log2asc`, and `log2long`
- Built the complete project with Make
- Verified the issue PoC under AddressSanitizer
- Confirmed unterminated input is rejected without an ASan error
- Smoke-tested valid CAN CC, CAN FD, CAN XL, and RTR inputs
- `git diff --check` passed

CMake configuration was not completed in the local WSL environment
because `pkg-config` is unavailable; the Make builds passed.

Fixes #632


hartkopp:
Nothing to fix here.
https://github.com/linux-can/can-utils/issues/632

